PR #15, opened 2019-04-29
Vim support - coc.nvim and ALE

antew:
This lets the language server work with Vim or ALE

Some notes on the implementation:

- I had to add a way to figure out whether to download the node or electron version of tree-sitter-elm, I think there has to be a better way to detect this but I haven't come up with anything.  I tried with the `is-electron` package but I think it always reports false when the server is started through vscode because it is really running in node at that point and communicating through ipc/stdio.
- When running with `coc.nvim` it wasn't as easy to get the `elmWorkspace` folder into the initialization options, but it does provide a `rootUri`, so I added that as a fallback when `elmWorkspace` is unavailable.
- I found that ALE doesn't support workspace folders, I saw it was only using the first folder in a workspace for now so I simplified the code a bit there to only use the workspace folder or the one taken from the `rootUri`

I updated the readme with info on how to enable it for the different editors.

Here's how the warnings appear in Vim with coc.vim

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,3 +11,36 @@ This vscode extension is in development and might be lacking features you know w
 This extension contributes the following settings:
 
 * `elmLS.trace.server`: enable/disable trace logging of client and server communication
+
+## Editor Support
+
+### Vim
+
+#### coc.nvim
+To enable support with [coc.nvim](https://github.com/neoclide/coc.nvim), run `:CocConfig` and add the language server config below.
+
+```
+{
+  "languageserver": {
+    "elm-ls": {
+      "command": "elm-ls",
+      "args": ["--stdio"],
+      "filetypes": ["elm"],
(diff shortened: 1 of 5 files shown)


Razzeee:
So the node builds are actually what you use for this? I just started building them some versions ago, for no particular reason. Would be good to know.

We will need more client parameters in the future, i'm trying to get around that, but I doubt we will really be able.

antew:
With the electron build it was working great in VSCode, but when I was trying to use it from Vim I kept getting a `Server elm-ls failed to start: Request initialize failed unexpectedly without providing any details.`.

It doesn't give a stack trace or any info really on what went wrong, I'll run it through a debugger tonight and see if I can some details on it.

I noticed the node packages in releases so I gave it a shot and it worked the first try!

Issue #13, opened 2019-04-25
diagnosticProviders


Razzeee:
We currently have two ways diagnostics are generated, there might be more in the future:

- elm make
- elm analyse

- [ ] Testing
- [ ] Bugfixes
- [x] Possibility to give the path to both if they're not installed global

Elm analyse works fine on windows for me.
Elm make tries to open the generated `elm.js` every time it runs. Pretty weird, hopefully just something that happens when you run the dev version.

Linux says it doesn't global installations, I do have them. So it might be my linux or a problem in general.
No idea about macOS.

antew:
`elm-analyse` wasn't working for me on macOS, I was using the elm-spa-example for testing, it gives me this error

```
Debugger listening on ws://127.0.0.1:6010/2f37d3a6-071e-4a11-b12a-240f204573d7
For help, see: https://nodejs.org/en/docs/inspector
[Info  - 5:19:33 PM] Rebuilding tree-sitter for local Electron version
[Info  - 5:19:33 PM] Fetching tree-sitter
[Info  - 5:19:33 PM] Downloading from https://github.com/tree-sitter/node-tree-sitter/releases/download/v0.14.0/tree-sitter-v0.14.0-electron-v64-darwin-x64.tar.gz
[Info  - 5:19:33 PM] Fetching tree-sitter-elm
[Info  - 5:19:33 PM] Downloading from https://github.com/razzeee/tree-sitter-elm/releases/download/v2.2.1/tree-sitter-elm-v2.2.1-electron-v64-darwin-x64.tar.gz
[Info  - 5:19:33 PM] Rebuild succeeded!
[Info  - 5:19:34 PM] Initializing Elm language server for workspace
 file:///Users/hpate/git/elm-spa-example using file:///Users/hpate/git/elm-spa-example/
[Info  - 5:19:34 PM] Reading elm.json from /Users/hpate/git/elm-spa-example/elm.json
[Info  - 5:19:34 PM] source-dirs found: /Users/hpate/git/elm-spa-example/src
[Info  - 5:19:34 PM] Found 33 files to add to the project
undefined:1
[...]

PR #16, opened 2019-04-29
Faster elm-analyse support

antew:
Updates how the diagnostics are run so that `elm-analyse` and `elm make` can run concurrently and report back with diagnostics when they finish, I saw your comment on #13 and I totally agree Rx could make for a nice API for subscribing to these.  I also plan to update elm-analyse to allow requesting a fix from the server as well as batch applying fixes on a file so it would be nice for that too!

This currently relies on a fork I have of elm-analyse that allows passing in the file content on changes rather than needing to read from disk, which allows linting as you type with it.  [elm-analyse#201](https://github.com/stil4m/elm-analyse/pull/201).

lint-as-you-type

### DIFF
--- a/server/src/providers/diagnostics/diagnosticsProvider.ts
+++ b/server/src/providers/diagnostics/diagnosticsProvider.ts
@@ -1,10 +1,9 @@
 import {
   Diagnostic,
   DiagnosticSeverity,
-  DidOpenTextDocumentParams,
-  DidSaveTextDocumentParams,
   IConnection,
-  PublishDiagnosticsParams,
+  TextDocuments,
+  TextDocumentChangeEvent,
   Range,
 } from "vscode-languageserver";
 import URI from "vscode-uri";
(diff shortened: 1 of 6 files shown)


Razzeee:
Very awesome.

Minor note, just to keep us talking about the same stuff.
> I also plan to update elm-analyse to allow requesting a fix from the server as well as batch applying fixes on a file so it would be nice for that too!

Would be awesome, but applying the fix should not be done by elm-analyse but by us. As there's an api for that. Best would be a diff (or well it's named editList I think) that we can feed to the lsp client, which then would execute it.

antew:
That sounds good! I'm with you on how elm-analyse should apply the fixes, right now it has to write the file to disk between each fix, but I'll would be much nicer if it could return a `TextEdit` so that it can be applied.

antew:
This should be good to go now :)

server/src/providers/diagnostics/diagnosticsProvider.ts
```diff
@@ -1,10 +1,9 @@
 import {
   Diagnostic,
   DiagnosticSeverity,
-  DidOpenTextDocumentParams,
-  DidSaveTextDocumentParams,
   IConnection,
-  PublishDiagnosticsParams,
+  TextDocuments,
```
Razzeee: Please install tslint, these need to be sorted down. And there are a few more warnings.

Razzeee:
This only seems to parse files we opened or changed. But we should get diagnostics for all files in the project, not just the ones that were opened.

If you don't feel like figuring that out, I can probably have a go at it.

server/src/providers/diagnostics/elmAnalyseDiagnostics.ts
```diff
@@ -1,123 +1,129 @@
...
-        messageInfoProperties.range2,
-      ];
-    } else if (messageInfoProperties.hasOwnProperty("ranges")) {
-      messageInfoFileRanges = messageInfoProperties.ranges;
-    } else {
-      messageInfoFileRanges = [[0, 0, 0, 0]];
-    }
-    return messageInfoFileRanges;
-  }
+    // We you fix the last error in a file it no longer shows up in the report, but
```
Razzeee: Typo

antew: 🤦‍♂ Fixed up with b2ce83f

server/src/providers/diagnostics/diagnosticsProvider.ts
antew: No problem, I cleaned up the warnings with 330d4fe

I noticed in the `tslintconfig.json` it didn't have a glob for the `src` files so I added that in, and for picking up the `.prettierrc.yaml` in 4e9ec8b

antew: Hmm, with the change in tsconfig it complaints about the project not being set to `composite: true` when building, I reverted it now for c27bb33 because it requires other changes to set that up

antew:
> This only seems to parse files we opened or changed. But we should get diagnostics for all files in the project, not just the ones that were opened.
> 
> If you don't feel like figuring that out, I can probably have a go at it.

I can check it out later tonight, I think it is a change I made in elm-analyse with allowing file content to be passed in.

antew:
> This only seems to parse files we opened or changed. But we should get diagnostics for all files in the project, not just the ones that were opened.
> 
> If you don't feel like figuring that out, I can probably have a go at it.

Fixed with f5d7305, though I am ashamed to admit how long it took me to debug it!

Razzeee:
This works wonderfully, thank you! :confetti_ball:

Razzeee:
This implementation is somehow causing the astProvider to not get document updates anymore. It normally puts a log line when you change a document, that's not happening anymore.

I tracked it down to this code adding `this.documents.listen(connection);`

antew:
Sorry about that, I'll take a look.  I think it is that these lines in the `diagnosticsProvider.ts` end up overwriting the listeners in the `astProvider.ts`
```
this.documents.onDidOpen(this.getDiagnostics);
this.documents.onDidChangeContent(this.getDiagnostics);
this.documents.onDidSave(this.getDiagnostics);
```

```
    this.connection.onDidChangeTextDocument(this.handleChangeTextDocument);
    this.connection.onDidCloseTextDocument(this.handleCloseTextDocument);
```

The listeners may have to be moved up a level so that they are shared.

Razzeee:
yeah, suspected that too, in fact had that problem some time ago with something else.

PR #20, opened 2019-05-06
Fix event listeners being overridden

antew:
Allows multiple event listeners for notification events in the document (open, close, save, change).

In vscode-languageserver the built in events for open, close, change, etc, can only have one handler attached, adding a new one overwrites the old one.  For requests that require a response from the language server that is only reasonable, it avoids needing to merge responses from multiple listeners and whatnot, but for notification events that don't require a response it is not as nice to work with.

A second problem this solves is allowing partial updates in the astProvider.  The TextDocuments class from vscode-languageserver almost fits the bill, but when a document changes it applies the changes directly and wouldn't allow tree-sitter to incrementally parse.

You can now listen on low level events with the `DocumentEvents` class, which is useful for the astProvider, and higher-level TextDocument events with `TextDocumentEvents`

### DIFF
--- a/server/src/providers/astProvider.ts
+++ b/server/src/providers/astProvider.ts
@@ -18,6 +18,7 @@ import {
 import URI from "vscode-uri";
 import { IForest } from "../forest";
 import { Position } from "../position";
+import { DocumentEvents } from '../util/documentEvents';
 import * as utils from "../util/elmUtils";
 
 export class ASTProvider {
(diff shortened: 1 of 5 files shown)


Razzeee:
I only seem to get elm analyse diagnostics for files I've opened?

Razzeee:
Hrm, started to work now, all of a sudden.

PR #23, opened 2019-05-08
elm-analyse: Fix cache directory location

### LINKED ISSUE #22: Elm-stuff folder always will be created at the root level
Doesn't seem right, should always be on the level of the elm.json

antew:
Previously it based the cache path off of the current working directory, which may not always be the same as the project root, it now uses the folder where elm.json is as the base for the cache path.

I updated the PR over in https://github.com/stil4m/elm-analyse/pull/201 as well.

Closes #22


Razzeee:
Closes #22

PR #24, opened 2019-05-12
Make elm and elm-format configurable. Compute editor input with elm f…

Razzeee:
…ormat not content from disk.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,15 +8,19 @@ This vscode extension is in development and might be lacking features you know w
 - Formatting via elm-format
 
 ## Extension Settings
+
 This extension contributes the following settings:
 
-* `elmLS.trace.server`: enable/disable trace logging of client and server communication
+- `elmLS.trace.server`: Enable/disable trace logging of client and server communication
+- `elmLS.elmPath`: The path to your elm executeable.
+- `elmLS.elmFormatPath`: The path to your elm-format executeable.
 
 ## Editor Support
 
 ### Vim
 
 #### coc.nvim
+
 To enable support with [coc.nvim](https://github.com/neoclide/coc.nvim), run `:CocConfig` and add the language server config below.
 
 ```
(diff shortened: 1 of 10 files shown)


Razzeee:
@antew Please see if this will create problems with VIM.
I hope it can also set these configs, otherwise we will need a fallback.

antew:
I was trying it out and I wasn't able to coc.vim to see the config yet, it has two options to pass config through, there's a `settings` key and a `initializationOptions` one.  I know the initialization options works, but it would be much nicer if it were unified.

I'll run it through the debugger and see if I can figure out where I'm going wrong with it.

bengolds:
@antew I just went through the whole process as well, and here's my understanding:
[This appears to be the function](https://github.com/neoclide/coc.nvim/blob/6efe8037a60e23f981cf0816e3d93af17f7710d4/src/services.ts#L287) that digests the data from `coc-settings.json`. It only pulls very specific configuration values. I don't think it even touches the `settings` key!

When you call `connections.workspace.getConfiguration()` from [the language server](https://github.com/elm-tooling/elm-language-server/blob/ab3b7e76014f685535380706407f1180658c29b0/server/src/util/settings.ts#L12), you get a very specific set of sections. As far as I can tell, there is no way to arbitrarily add sections to this config, so we could never have `elmLS` at the top level. Anything inside of the language server's `settings` key from `coc-settings` also doesn't appear in here at all.
ScreenCapture at Wed May 15 12:42:36 PDT 2019

All this is telling me that I think `initializationOptions` is the right place for this.

bengolds:
Alright, after more research, it appears that the correct way to do this is to [write a coc extension](https://github.com/neoclide/coc.nvim/wiki/Using-coc-extensions).

As an example, [here is the rust extension](https://github.com/neoclide/coc-rls).

antew:
Thank you for doing the research @bengolds!

I think `initializationOptions` is the nicest way to go about it given the available options, I checked ALE and it supports them too.

The coc forks make sense for turning a VSCode extension into a nicer Vim experience because you can build off the existing extension, maybe the ideal place for the Vim configuration with the server would be in https://github.com/ElmCast/elm-vim? I don't know if the coc setup has to be a fork, or if it is only required if you want to convert a VSCode extension to a Vim one.

antew:
It looks like it can be set with coc through VimL

```
call coc#config("elmLS", { 'elmPath': 'path/to/elm-format' })
```

image

Razzeee:
Does that mean I can merge this?

antew:
Merging sounds good to me, I can put in a PR to add ALE support for those variables in the next day or two.

Issue #29, opened 2019-05-17
Errors don't disappear when last error is from elm-make


bengolds:
Repro steps:
1. Get a file to have only one `elm-make` error and no `elm-analyse` warning.
2. Fix the `elm-make` error.

Expected:
No errors reported.

Observed:
Last elm-make error still appears in diagnostics.

This doesn't happen if the last error is an `elm-analyse` warning.

(Note: observed in both VSCode and vim)

Razzeee:
Should be addressed now.

PR #31, opened 2019-05-18
Clear make diagnostics on file when no errors are left (#29)

bengolds:
This should fix issue #29. I don't think there are any other consequences to doing this, although it might change if we start running `elm-make` on unsaved buffers.

### DIFF
--- a/src/providers/diagnostics/diagnosticsProvider.ts
+++ b/src/providers/diagnostics/diagnosticsProvider.ts
@@ -99,27 +99,7 @@ export class DiagnosticsProvider {
 
     this.elmAnalyseDiagnostics.updateFile(uri, text);
 
-    const compilerErrors: IElmIssue[] = [];
-    compilerErrors.push(
-      ...(await this.elmMakeDiagnostics.createDiagnostics(uri)),
-    );
-
-    const diagnostics: Map<string, Diagnostic[]> = compilerErrors.reduce(
-      (acc, issue) => {
-        // If provided path is relative, make it absolute
-        if (issue.file.startsWith(".")) {
-          issue.file = this.elmWorkspaceFolder + issue.file.slice(1);
-        }
-        const issueUri = URI.file(issue.file).toString();
-        const arr = acc.get(issueUri) || [];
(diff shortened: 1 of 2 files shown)


Razzeee:
Is this the same as this handles?
https://github.com/elm-tooling/elm-language-server/blob/master/server/src/providers/diagnostics/elmAnalyseDiagnostics.ts#L95

If yes, we might want to drop that and fix it like you did.
@antew

bengolds:
Correct — the only benefit I can think of to the way @antew did it was that it works for files other than the one you’re looking at. That doesn’t matter in my environment (only shows errors on current file) but it could in others and is probably desirable, now that i think about it. 

I thought about applying that approach to elm make, however it’s a bit more challenging to keep track of when other files are clean: elm-analyse always runs across the whole project, while elm make only runs across the current file and its indirect dependencies. 

If we could get a list of all the files elm-make ran over and succeeded on, I think @antews approach is better.

Razzeee:
I'm surprised by your statement. I think that elm-make runs always on the whole solution. So if it returns, we can safely drop all previous elm make errors. Even if we got an empty result.

bengolds:
[elm-make-test.zip](https://github.com/elm-tooling/elm-language-server/files/3194241/elm-make-test.zip)
Doesn't look like it, sadly. When you run `elm-make`, you give it a file to run on. It then compiles that file and anything it depends on. Try running `elm-make` on the 4 different files in that test package, and you'll see what I mean.

For now, my approach is probably an okay stop-gap. Realistically, we probably want to figure out the dependency graph for a file, and on success, mark _all_ of the files it depends on as fixed.

Razzeee:
Can you check if we can move this into the elmMakeProvider specific code then?

bengolds:
Finally got around to moving this into `elmMakeProvider` 👍

bengolds:
Annnnd it needs more testing. Hold, please 😄 ☎️

bengolds:
Just needed to fix a small typo. After using this all day, it seems okay 👍

Razzeee:
Thank you!

PR #35, opened 2019-05-22
Add fallbacks for workspace/configuration settings

antew:
This allows falling back to getting the `elmPath` and `elmFormatPath` from the `initializationOptions` so that support for coc.vim and ALE is seamless.

I updated the readme with new instructions for coc.vim, and pushed a new version of https://github.com/antew/vim-elm-language-server that allows configuring the paths with ALE.

I had to add a capability check on `workspace/configuration` for ALE, the promise would never resolve or reject with clients that didn't respond to the message, so there is a short-circuit for that case.


Razzeee:
Thank you 🎊

PR #40, opened 2019-05-27
Define "bin" for "npm link" to work

andys8:
The user still needs to execute the compile step before. `npm link` will link the output and make it executable under the given name.

Note: This will/would use the name `elm-language-server`. The plugins are using `elm-lsp` and the `README.md` has `elm-ls` in it. We choose a name and fix it. Your choice @Razzeee  ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ If needed, you can set the paths to `elm` and `elm-format` with the `elmPath` an
 {
   "languageserver": {
     "elmLS": {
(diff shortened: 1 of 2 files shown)


Razzeee:
Makes sense, as I separated client and server yesterday and the client has this in the package.json https://github.com/elm-tooling/elm-language-client-vscode/blob/master/package.json#L29

Razzeee:
Can you align the readme too?

PR #41, opened 2019-05-30
Fix global elm analyse result being replaced by the result of a singl…

Razzeee:
…e file

Before this, if you initialize your workspace you will only get elm-analyse results for the current file, as the result the server pushes will get overwritten.

@antew 
not sure this is the ideal fix, but the problem seems to be gone afterwards.


antew:
Yeah this is related to that timing bug with the `elm-analyse.json`, the file change event is kicking elm analyse out of its normal state machine flow.  I'd like to make it wait for the first report to come back from elm-analyse, I put in #42 for it but I need to test it out still.

Issue #30, opened 2019-05-17
`elm-analyse.json` is getting ignored


bengolds:
I haven't been able to fully track down why, but settings in `elm-analyse.json` are getting ignored. As an example, if you have:
```
{
    "checks": {
        "ImportAll": false
    },
}
```
`elm-ls` still reports `ImportAll` warnings. 

From my digging, it looks like the port that loads `elm-analyse.json` is getting initialized correctly:
From [file-loading-ports](https://github.com/stil4m/elm-analyse/blob/455a42cba07012a67f498bca106d35688a179c95/ts/file-loading-ports.ts#L14) to [context](https://github.com/stil4m/elm-analyse/blob/455a42cba07012a67f498bca106d35688a179c95/ts/ports/context.ts).

I haven't poked the internals of elm-analyse to understand why it's ignoring them.

Razzeee:
@antew can you have a look at one point?

antew:
Yeah no problem, I'm on vacation until tomorrow but I'll be back tomorrow night and can check it out.

Razzeee:
Take your time please, vacations are sacred :palm_tree:

antew:
It looks like there are two bugs in how the `elm-analyse.json` file is handled.

1. I inadvertently introduced a timing bug in elm-analyse, if the first file change event happens in the language server before `elm-analyse` has loaded the configuration file it will skip the state where it is listening for the configuration and go with the default.
2. It would always try to load the configuration relative to the current working directory rather than the directory where `elm.json` is.

I'm still debating how to fix them, for `elm-analyse.json` I am leaning toward having it look up through parent directories so that you have the option of setting global configuration in `~/elm-analyse.json`.

For the timing bug the fix is either going to be passing in the configuration as a `flag`, or ignoring the file change messages until everything has loaded, I am leaning toward the `flag` approach at the moment because it seems cleaner and would make it easier to add Elm Analyse settings panes and whatnot later on.

I should be able to get in a PR for it tomorrow night.

antew:
For the timing bug #42 should fix it, and for multi-projects where you might have `/project` open in an editor and the elm app is at `/project/web/` I put in a PR for elm-analyse over at https://github.com/stil4m/elm-analyse/pull/204

Cheers!

andys8:
I had trouble before, but it's now working for me with a simple project structure where `elm.json` and `elm-analyse.json` are located in the project / git root.

Razzeee:
[...]

PR #42, opened 2019-05-31
WIP: Fix for making sure elm-analyse is initialized before we send updates

antew:
Partial fix for #30, and fixes an issue with results for the project being overwritten by the results for a single file on startup.

I still need to finish setting up the new project to test it out, I tested it on the previous version with the client/server combined.


Razzeee:
I tested this and it fixes the problem with overwriting diagnostics.
I did not test the elm-analyse.json

hpate-omicron:
On my work account now but it should fix the `elm-analyse.json` being overwritten by the default config then as well, it'll give it time to load the user config first before sending over any file change events.

For the issue with a multi-project structure I opened a PR over on elm-analyse for it.

On another note I updated the executable name in https://github.com/antew/vim-elm-language-server so it is up to date now as well.

I'll put in a PR in the near feature to add it into ALE itself, that way if someone has `elm-language-server` installed and `ALE` installed for Vim it will pick it up automatically and start using it.

PR #46, opened 2019-06-01
Use wasm for language server

### LINKED ISSUE #44: Use web assembly version of tree sitter
If we use the web assembly version, we can probably get rid of rebuilder and the needed download logic.

https://github.com/georgewfraser/vscode-tree-sitter can be used as a reference on how to do this.

Razzeee:
Closes #44


Razzeee:
https://github.com/tree-sitter/tree-sitter/issues/349

PR #55, opened 2019-06-09
Elm-analyse code actions

antew:
I think this is about ready now, it adds in a code action provider that supports elm-analyse.  
The upstream set of changes for `elm-analyse` is https://github.com/stil4m/elm-analyse/pull/209.

fixer

Features:
- `elm-analyse` messages that can be fixed now show up as code actions in the editor!
- `elm-format` is run on the fixed text

Changes:
- Moved common diffing code into `src/util/diff.ts`, it's now used by the `documentFormattingProvder` and `elmAnalyseDiagnostics` to generate changelists for the document so I moved it over there and have both importing it.
- Moved `elmAnalyseDiagnostics` up a level in the code so that it can be given to the `diagnosticsProvider` and the `codeActionProvider`.
- Broke out a block of code in the `documentFormattingProvider` into a public function for formatting some text, this way elm-analyse can make use of `elm-format` as well and we can apply the already formatted fix to the document.

Misc:
- I noticed that the `TextDocumentEvents` was importing `EventEmitter` from `ws`, so I switched it to use it from `events` like the other files.

### DIFF
--- a/src/providers/diagnostics/elmAnalyseDiagnostics.ts
+++ b/src/providers/diagnostics/elmAnalyseDiagnostics.ts
@@ -33,7 +33,7 @@ const fixableErrors = [
   "DuplicateImport",
 ];
 const ELM_ANALYSE = "elm-analyse";
-export const CODE_ACTION_ELM_ANALYSE = "elmLanguageServer.elmAnalyseFixer";
+export const CODE_ACTION_ELM_ANALYSE = "elmLS.elmAnalyseFixer";
 
 export interface IElmAnalyseEvents {
   on(event: "new-report", diagnostics: Map<string, Diagnostic[]>): this;
--- a/src/providers/diagnostics/elmAnalyseDiagnostics.ts
+++ b/src/providers/diagnostics/elmAnalyseDiagnostics.ts
@@ -11,10 +11,13 @@ import {
   DiagnosticSeverity,
   ExecuteCommandParams,
   IConnection,
+  TextDocument,
 } from "vscode-languageserver";
 import URI from "vscode-uri";
 import * as Diff from "../../util/diff";
+import { Settings } from "../../util/settings";
 import { TextDocumentEvents } from "../../util/textDocumentEvents";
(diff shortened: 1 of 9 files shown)


src/providers/diagnostics/elmAnalyseDiagnostics.ts
```diff
@@ -1,32 +1,68 @@
...
+  "UnusedImportedVariable",
+  "UnusedImportAlias",
+  "UnusedPatternVariable",
+  "UnusedTypeAlias",
+  "MultiLineRecordFormatting",
+  "DropConsOfItemAndList",
+  "DuplicateImport",
+];
+const ELM_ANALYSE = "elm-analyse";
+export const CODE_ACTION_ELM_ANALYSE = "elm-analyse-fixer";
```
Razzeee: As you seem to be using this for the command name, I think we should try to mimic the build in command names like these https://code.visualstudio.com/api/references/commands

Razzeee:
Thank you. Code looks good, just a minor.

Will also need a new elm-analyse version.

src/providers/diagnostics/elmAnalyseDiagnostics.ts
antew: Sounds good! I changed it over to `elmLanguageServer.elmAnalyseFixer` with 38e7490, if you thing something else work work better let me know.

antew:
I pushed a branch with the built `elm-analyse` files so that it is easier to test out, it's available iwth `npm install "antew/elm-analyse#lsp-fixer-built" --save`

src/providers/diagnostics/elmAnalyseDiagnostics.ts
Razzeee: I think we settled for `elmLS` instead of `elmLanguageServer`

antew: Sweet, updated w/398ef5f

PR #62, opened 2019-06-13
kakoune capabilities table

bburdette:
Ok I added a bunch of checks to the capabilities table, with some exceptions:

codeLenses, I didn't see what looked like a relevant command.  
fold - no command with fold in the name is visible from kakoune.

workspace-symbols has a command present, but I don't know how to make it do anything, or what its supposed to do.  document-symbols appears to work though, so maybe workspace does too?  unchecked for now.


Razzeee:
Thank you!

Workspace symbols are the same as document symbols, but not document bound. Means you can find stuff from other files.

Fold doesn't need to be a command, but just being able to fold the code in the editor. Then again, that might be a concept foreign to kakoune (never used it).

Codelenses should show up automatically for every function, if your client supports them.

Issue #36, opened 2019-05-26
Build and lint with travis


Razzeee:

buinauskas:
I'd like to help with this if you're willing to give a more details.

Razzeee:
Well I started with here https://github.com/elm-tooling/elm-language-server/pull/37

But I think it's failnig due the the navtive builds and me not including clang etc.  like I do here https://github.com/Razzeee/tree-sitter-elm/blob/master/.travis.yml

It should get way easier when we are able to switch to the wasm version of tree sitter. But it's blocked until somebody tackles https://github.com/tree-sitter/tree-sitter/issues/349

PR #65, opened 2019-06-17
Adding Travis configuration

buinauskas:
@Razzeee Hey, I've picked up #36 and have made the following changes:
-   Added Travis configuration
-   Configured pre-commit hook to automatically format staged code
-   Configured pre-commit hook to automatically lint files
-   Fixed linting issues project had (very minor ones)
-   Added CI build step to check if code is actually formatted as per spec
-   Changed few typos (Will submit another PR to fix more of these)

I've never worked with Travis CI before, so apologies if I did something incorrectly - I'd be glad if I can make this work. I've tested this on my fork and build was successful. Let me know if there's something I can improve, thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: node_js
+
+node_js: 10
+
+env:
+  - CXX=clang++
+
+branches:
+  only:
+    - master
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,16 @@ language: node_js
 node_js:
   - "stable"
 
+sudo: required
+before_install:
+  - npm install -g node-gyp
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+    packages:
+      - g++-5
+
 install:
   - npm ci
 cache:
@@ -10,7 +20,7 @@ cache:
     - "$HOME/.npm"
 
 env:
-  - CXX=clang++
+  - CXX=g++-5
 
 branches:
   only:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ language: node_js
 node_js:
   - "stable"
 
+install:
+  - npm ci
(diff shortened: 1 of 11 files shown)


.travis.yml
```diff
@@ -0,0 +1,16 @@
+language: node_js
+
+env:
+  - CXX=clang++
+
+branches:
+  only:
+    - master
```
Razzeee: Hrm, I think we should build on all branches. That way we can see if something is wrong, before it get's merged :)

buinauskas: Sure, totally makes sense :+1:  Removed restriction.

package.json
```diff
@@ -42,9 +45,16 @@
   "scripts": {
     "compile": "tsc -b",
     "watch": "tsc -b -w",
+    "format:check": "prettier --config ./.prettierrc.yaml --list-different \"src/**/*{.ts,.js,.json}\"",
```
Razzeee: Do we need this? Shouldn't "lint" report these too, as we're using tslint-plugin-prettier?

Razzeee:
Thanks for taking the time to do this. I just found some minors, after that  it's ready to go in.

package.json
buinauskas: I've always used this kind of approach, but I'm happy to check if tslint-plugin-prettier does exactly the same. Will update based on my findings :)

buinauskas: Reverted.

buinauskas: Removed redundant check :+1:

.travis.yml
buinauskas: This is rather silly, but should I resolve this conversation upon resolving issues or it's better if you'd check that and close it?

Razzeee: I like to have others check and resolve these discussions :)

Razzeee: So in this case me ;)

PR #67, opened 2019-06-18
Configurable if diagnostics run on change

andys8:
Settings `diagnosticsOn` is set to "change" by default. Otherwise it will not register the change handler for diagnostics. I'm open for a different name or type. I went with string as a value, because I'm not sure about types in vim or other editor configurations.

### DIFF
--- a/src/util/settings.ts
+++ b/src/util/settings.ts
@@ -4,10 +4,12 @@ export interface IClientSettings {
   elmFormatPath: string;
   elmPath: string;
   elmTestPath: string;
+  diagnosticsOn: "change" | "save";
 }
 
 export class Settings {
   private fallbackClientSettings: IClientSettings = {
+    diagnosticsOn: "change",
     elmFormatPath: "elm-format",
     elmPath: "elm",
     elmTestPath: "elm-test",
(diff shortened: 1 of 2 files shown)


src/util/settings.ts
```diff
@@ -24,23 +26,21 @@ export class Settings {
...
       this.capabilities.workspace.configuration;
 
+    const defaultSettings = {
+      ...this.fallbackClientSettings,
+      ...this.initializationOptions,
+    };
+
     if (!supportsConfig) {
-      return Promise.resolve(this.initializationOptions);
+      return Promise.resolve(defaultSettings);
```
andys8: @Razzeee I think not including `fallbackClientSettings` here before could have been a bug.

Razzeee:
String should be fine.

I think we should actually run both elm make and elm-analyse on the same event. And at least try if we can default to on change.

andys8:
@Razzeee Because of recent changes like not running `elm-analyse` when `elm make` fails, and the behavior in coc and ale (which turned out to be configurable), I'd suggest to not merge this at the moment. Maybe it makes sense to have configurable options, but for now I'd close this PR if you don't mind.

Razzeee:
@andys8 can you please PR the fallback settings then?

andys8:
Sure https://github.com/elm-tooling/elm-language-server/pull/78

andys8:
In Vim / Coc this will set the setting to diagnostics on save only.
```
"diagnostic.refreshAfterSave": true,
"diagnostic.refreshOnInsertMode": false,
```

In Vim / Ale it is:

```
let g:ale_lint_on_text_changed = 0
let g:ale_lint_on_insert_leave = 0
let g:ale_lint_on_save = 1
```

We could add this to the README.md

Issue #70, opened 2019-06-18
Repetetive types


buinauskas:
There's quite a lot of repetitive types that are inlined within generic types, such as arrays.

For instance:
``` typescript
private findElmFilesInFolders(
  elmFolders: Array<{ path: string; writable: boolean }>,
): Array<{ path: string; writable: boolean }> {
  let elmFilePaths: Array<{ path: string; writable: boolean }> = [];
  for (const element of elmFolders) {
    elmFilePaths = elmFilePaths.concat(this.findElmFilesInFolder(element));
  }
  return elmFilePaths;
}

private findElmFilesInFolder(element: {
  path: string;
  writable: boolean;
}): Array<{ path: string; writable: boolean }> {
  return glob
    .sync(`${element.path}/**/*.elm`)
    .map(a => ({ path: a, writable: element.writable }));
}
```

The following type appears 4 times in snippet above.
``` typescript
{ path: string; writable: boolean }
```

This could be extracted into a separate interface. This will improve readability and reduce code duplication.

Razzeee:
There are a bunch of these, as I did not take the time to find names. Cause naming is hard :)
Please feel free to improve them, but I'm closing this issue due to the open endedness of this task.

Issue #69, opened 2019-06-18
Enforce consistent string concatenation


buinauskas:
At the moment strings are concatenated in two ways:
-   `+` operator
-   using template literals

I'd say code should support just one of these for the sake of consistency and enforce that using https://palantir.github.io/tslint/rules/prefer-template/ to allow just template literals.

PR #71, opened 2019-06-18
Consistent string concatenation, IFolder interface instead of inlined types

buinauskas:
- Added missing syntax highlighting in README
- Consistent string concatenation with TSLint rule for it
- Added `IFolder` interface to extract repeated type.

This addresses #69 and #70


Razzeee:
Thanks for those awesome changes!

Issue #61, opened 2019-06-13
Describe CoC Feature set


Razzeee:
elm-tooling/elm-language-server#editor-support

Issue #60, opened 2019-06-13
Describe ALE Feature set


Razzeee:
https://github.com/elm-tooling/elm-language-server#editor-support

PR #72, opened 2019-06-19
Add some documentation on what Coc and ALE support.

antew:
For #60 and #61

Coc actually supports more than I realized, I didn't know it had codelens support if you have neovim.


Razzeee:
Awesome, thank you

PR #74, opened 2019-06-20
Add support for fixing all elm-analyse issues in a file

antew:
To go along with https://github.com/elm-tooling/elm-analyse/pull/6

fix-example

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -847,7 +847,7 @@
       "integrity": "sha1-WQxhFWsK4vTwJVcyoViyZrxWsh0="
     },
     "elm-analyse": {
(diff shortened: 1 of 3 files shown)


Razzeee:
This is WIP due to https://github.com/elm-tooling/elm-analyse/pull/6 not being merged or anything else?

antew:
It was WIP due to the elm-analyse piece, it should be good to go once I update it to point to the jew elm-analyse version.

antew:
Added a PR over on the elm-analyse fork with the built files https://github.com/elm-tooling/elm-analyse/pull/7

It feels weird how elm-analyse has the built files in master, maybe we should create branches off of master, build the files there, and tag those branches instead?

Razzeee:
Hrm, I'm fine for the moment with how it is. The upstream repo does it the same way.

PR #78, opened 2019-06-21
Fix: Use fallbackSettings when workspace config is not supported

andys8:
`fallbackClientSettings` were previously ignored in this case.


Razzeee:
Thank you 🎊

Issue #89, opened 2019-06-23
Heavy CPU load when typing


andys8:
I'm in Insert Mode, typing and removing characters, not even leaving insert mode, and `elm-test` is fired multiple times.

This is in a test file. There is increased CPU also in source files, but not as bad. This could be because `elm` is better performing than `elm-test` or there is another difference.

![image](https://user-images.githubusercontent.com/13085980/59976016-327f3380-95bf-11e9-8302-13049a5ea6a1.png)

### Config

```
{
  "diagnostic.refreshAfterSave": true,
  "coc.preferences.formatOnSaveFiletypes": ["elm"],
  "coc.preferences.snippets.enable": false,
  "suggest.minTriggerInputLength": 3,
  "suggest.maxCompleteItemCount": 5,
  "diagnostic.refreshOnInsertMode": false,
  "languageserver": {
    "elmLS": {
      "command": "elm-language-server",
      "args": ["--stdio"],
      "filetypes": ["elm"],
      "rootPatterns": ["elm.json"],
      "trace.server": "verbose",
      "initializationOptions": {
        "runtime": "node",
        "elmPath": "elm",
        "elmFormatPath": "elm-format",
        "elmTestPath": "elm-test"
      }
    }
  }
}
```

LSP 1.0.2 (published), Vim / Coc

### Logs (partial)
```
[...]

[Trace - 2:04:02 PM] Received request 'workspace/configuration - (36)'.
Params: {
    "items": [
        {
            "section": "elmLS"
        }
    ]
}


[Trace - 2:04:02 PM] Sending response 'workspace/configuration - (36)'. Processing request took 0ms
Result: [
    {}
]


[Trace - 2:04:02 PM] Received notification 'window/logMessage'.
Params: {
    "type": 4,
    "message": "Received new elm-analyse report with 0 messages"
}


Received new elm-analyse report with 0 messages
[Trace - 2:04:02 PM] Received notification 'textDocument/publishDiagnostics'.
Params: {
    "uri": "file:///home/as/Dev/repository/vim-emulation/tests/MainTest.elm",
    "diagnostics": []
}


[Trace - 2:04:02 PM] Received notification 'textDocument/publishDiagnostics'.
Params: {
    "uri": "file:///home/as/Dev/repository/vim-emulation/tests/MainTest.elm",
    "diagnostics": []
}


[Trace - 2:04:02 PM] Received notification 'window/logMessage'.
Params: {
    "type": 3,
[...]

PR #94, opened 2019-06-26
Bugfix #89 elm test compiles tests only

andys8:
This PR will add working linting in tests and *not* run the tests, instead compile them only.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,9 +25,9 @@
       }
     },
     "@types/caseless": {
-      "version": "0.12.1",
-      "resolved": "https://registry.npmjs.org/@types/caseless/-/caseless-0.12.1.tgz",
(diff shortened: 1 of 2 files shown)


Razzeee:
Don't we want both? And as far as I could tell, the old test code did not work. So it might be better to redo it from scratch.

andys8:
A test can take 10 seconds and much more. I think executing test suites is a completely different feature than linting and the integration and diffing because of the performance impact is not easy at all. It can't just be randomly triggered.

andys8:
Updated package-lock because that seems to be the issue with CI.

andys8:
> And as far as I could tell, the old test code did not work. So it might be better to redo it from scratch.

Razzeee:
I reworked it in master and I'm getting make, tests and diagnostics with that in vscode.

Razzeee:
And yes, tests might be better as a command, haven't really looked into that api yet.

andys8:
I tested the changes made to master, and it seems like the issues were addressed. It's really looking good now :) Only `package-lock.json` has to be updated (CI is failing too).

Issue #58, opened 2019-06-12
Working with Yarn?


romariolopezc:
Hello

I ran all the commands from the readme with yarn and I get:
```sh
rlopez ~/R/p/elm-language-server *master λ y link
yarn link v1.16.0
success Registered "@elm-tooling/elm-language-server".
info You can now run `yarn link "@elm-tooling/elm-language-server"` in the projects where you want to use this package and it will be used instead.
Done in 0.07s.
```

And using: `elm-language-server --stdio` gives:
```sh
fish: Unknown command elm-language-server
```

Razzeee:
If you don't need to run master, you can just use the version from npm and
do a global install. The guides probably need updating.

Razzeee:
Can we close this?

PR #95, opened 2019-06-27
Fix for yarn global install

antew:
I was talking with @romariolopezc on slack and he found that the install worked with npm but not with yarn, I took a look and it was happening because with yarn's global installs the tree-sitter and tree-sitter-elm packages can be at the root of the global node_modules folder rather than inside the elm-language-server/node_modules/ folder.

I updated it to use `require.resolve` to find the tree-sitter and tree-sitter-elm package directories so that we can use node's module resolution to find it.

This should be the fix for the root cause of #58

### DIFF
--- a/src/util/rebuilder.ts
+++ b/src/util/rebuilder.ts
@@ -31,9 +31,10 @@ function fetchPrebuild(
 ): Promise<void | Error> {
   // tslint:disable-next-line: no-console
   console.info(`Fetching ${name}`);
-  const pkgRoot: string = path.resolve(
-    path.join(__dirname, "../../node_modules", name),
-  );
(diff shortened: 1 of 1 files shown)


Razzeee:
We can't test this without a release right?

antew:
I think it does require a release, when installing with yarn locally it puts them in your local node modules folder so the issue doesn't happen, I'm not sure if there is a way to simulate how it can install when installed globally from npm.

Razzeee:
I already pushed a new version so we will see I guess :)

romariolopezc:
I'm going to test this later today :) I'll post my results here

romariolopezc:
Tested, installed with yarn global and is working as npm :)

PR #98, opened 2019-06-29
Generate unique IDs for each fixer so that they can be spawned multiple times

antew:
This should be the fix for that 7hoenix was seeing on Slack, with a message like

```
[Info  - 8:47:19 AM] Done parsing all files.
[Error - 8:47:19 AM] Server initialization failed.
Error: command 'elmLS.elmAnalyseFixer' already exists
    at d.registerCommand (/Applications/Visual Studio Code 2.app/Contents/Resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:497:447)
    at Object.registerCommand (/Applications/Visual Studio Code 2.app/Contents/Resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:716:422)
```

I didn't realize that they had to be unique per instance! https://github.com/Microsoft/vscode-languageserver-node/issues/333

### DIFF
--- a/src/providers/diagnostics/elmAnalyseDiagnostics.ts
+++ b/src/providers/diagnostics/elmAnalyseDiagnostics.ts
@@ -1,3 +1,4 @@
+import * as crypto from "crypto";
 import { ElmApp, FixedFile, Message, Report } from "elm-analyse/ts/domain";
 import { EventEmitter } from "events";
 import * as fs from "fs";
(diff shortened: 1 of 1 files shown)


antew:
I haven't found a way to reproduce the error yet, but I believe this will fix it.

buinauskas:
@antew this is a bit random question, but instead of using `crypto` to generate Ids, would it make sense to use `Symbol` instead? They're supposed to be completely unique.

antew:
@buinauskas I think there would need to be a way to get a unique string representation out of the `Symbol`, the `CodeActionKind` needs to be a string for the server https://github.com/microsoft/vscode-languageserver-node/blob/58551741af1d90dd96d9c8a5b0069619884d8f22/types/src/main.ts#L2058

buinauskas:
Gotcha. Thanks for explaining! : )

PR #107, opened 2019-07-16
don't download the definition files if they're not needed

BrianHicks:
`prebuildInstall.download` will always try and unpack whatever you give it but cache the archive. This causes problems when packaging using immutable structures like Nix/NixOS; the `node_modules` directory ends up being read-only. To get around this, we look for a file (actually a directory in this case) that we know to be present in the unpacked files on disk.


Razzeee:
Thank you for sending this patch :confetti_ball:

PR #110, opened 2019-07-18
Create quickfix for module name typos

Razzeee:


andys8:
Nice! 

One thing. I would be cautious with depending on the "visualization" `#`. This is coupling the representation to internal implementation.

Razzeee:
Fully aware. We'll change it if we ever get colored messages

PR #115, opened 2019-07-21
Cleaned up forest.ts, fixed typo

buinauskas:
Getting myself more familiar with the code by doing minor tweaks.

Here conditionals weren't really needed for `find` statements because they already bring back `T | undefined`, so they can be brought back directly.

Removed redundant array initialization from constructor, because it already has default value assigned.

Fixed typo.

Using built in method to check if object is array, rather than using `instanceof`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## 1.2.1
+
+- Revert determination of used elm compiler version, as it was causing file open to go unnoticed
+
 ## 1.2.0
 
 - Use WASM version of tree-sitter and updated tree-sitter - This mean multiple parsing improvements
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@elm-tooling/elm-language-server",
-  "version": "1.1.1",
+  "version": "1.2.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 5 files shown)


src/util/elmUtils.ts
```diff
@@ -187,9 +185,8 @@ export async function getElmVersion(
     connection,
   );
 
-  const version = result.stdout
-    .substring(0, result.stdout.indexOf("\n"))
-    .trim();
+  const version = result.stdout.trim();
```
Razzeee: Please don't change this.

Razzeee:
And please also remove the unnecessary merge commit you included

src/util/elmUtils.ts
buinauskas: I'll revert it, but I'd like to understand reasoning behind this.

```
elm --version
```

returns just a single line, e.g.
```
0.19.0

```

so I thought trimming was enough? Or are there some extra lines that could be brought back?

Razzeee: Well, did you actually test it or just look at it, I'm pretty sure, that  I needed it to make it work.
If you test  it and can confirm, that it works without the substring, I'm willing to change this :)

buinauskas: I did check only what `elm - - version` outputs in my terminal. So I didn't really test it. I'll do and will come back! 👍

PR #116, opened 2019-07-22
Cleaned up forest.ts, fixed typo

buinauskas:
Getting myself more familiar with the code by doing minor tweaks.

Here conditionals weren't really needed for `find` statements because they already bring back `T | undefined`, so they can be brought back directly.

Removed redundant array initialization from constructor, because it already has default value assigned.

Fixed typo.

Using built in method to check if object is array, rather than using `instanceof`.

Recreating this in favor for #115


Razzeee:
Btw, if you had just force pushed to your old branch, your PR would have updated.

Issue #119, opened 2019-07-28
Document setup with LanguageClient-neovim


lpil:
It'd be great to have documentation on usage with https://github.com/autozimu/LanguageClient-neovim

I tried this configuration but received an error

```viml
let g:LanguageClient_serverCommands = {
      \ 'elm': ['elm-language-server', '--stdio'],
      \ }
```

![Screenshot 2019-07-28 at 13 40 04](https://user-images.githubusercontent.com/6134406/62006778-3877af80-b13d-11e9-9bf0-3e4a3b240fec.png)

Razzeee:
Against which version of the server did you try?

lpil:
Version 1.2.1, thank you

andys8:
Hmm, the documentation doesn't give a lot of details in the example. https://github.com/autozimu/LanguageClient-neovim/blob/next/INSTALL.md#5-configure-this-plugin

I would try it with `elm-language-server` 1.1.1, but that's a guess. The configuration doesn't set paths for the necessary binaries (`elm-format`) and others. So it could be that those need to be set (but there should be defaults). It could be that a binary is missing in the path. But with this error, no existing example configuration and no useful error message, it's hard to tell.

Would be nice if the result of this issue would be a working configuration and some tipps for other users.

lpil:
I have elm and elm-format on my path, that should be enough?

Here's an example config for some other language servers

```viml
let g:LanguageClient_serverCommands = {
      \ 'rust': ['rustup', 'run', 'stable', 'rls'],
      \ 'python': ['pyls'],
      \ 'go': ['gopls']
      \ }
```

andys8:
Documentation says `elm-test`, too. Not sure what happens when you don't have tests.

Today was a new version released: 1.2.2

Razzeee:
Does this work for you now?

lpil:
After updating elm-language-server I no longer get an error at start but when I attempt to do something like look up a type definition I get an error message `[LC] sending on a disconnected channel`.

Razzeee:
Any more logs?

lpil:
Here's the logs from me starting nvim, opening an elm file in an elm project and attempting to look up a type definition

```
#######
LanguageClient 0.1.146 ec4af74cc389f802c3c43a660c963049848a8a93
#######
[...]

PR #122, opened 2019-08-04
Avoid error when initializationOptions is not set

jonmast:
[The spec](https://microsoft.github.io/language-server-protocol/specification#initialize) defines this as an optional field, so we shouldn't rely on it being set. This causes problems for https://github.com/autozimu/LanguageClient-neovim since it doesn't set the option by default. This PR changes it to default to an empty object when undefined to avoid this issue.

Possible fix for https://github.com/elm-tooling/elm-language-server/issues/119, this change was sufficient for me to get it working with the following config, but I didn't see the exact error described there.

```vim
let g:LanguageClient_serverCommands = {
      \ 'elm': [ 'elm-language-server', '--stdio' ],
\ }

```

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -66,13 +66,16 @@ export class Server implements ILanguageServer {
     const elmWorkspaceFallback =
       // Add a trailing slash if not present
       this.params.rootUri && this.params.rootUri.replace(/\/?$/, "/");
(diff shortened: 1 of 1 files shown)


Razzeee:
Thank you

PR #126, opened 2019-08-10
Initialization changed

### LINKED ISSUE #125: Formatting fails when server isn't done initializing
Opening an editor starts the language server. It seems to trigger initial diagnostics. The one that can be perceived visually, is that it takes some time until `elm-analyse` warnings start showing up. Until this is the case, format on save seems to fail due to timeouts.

## Expected Behavior

(if possible) Format and save the file independent of `elm-analyse` diagnostics

## Current Behavior

Formatting fails until the server is initialized.

### Demo: Timeout until initialization
vokoscreen-2019-08-10_13-49-17

### Demo: Broken result
vokoscreen-2019-08-10_13-48-12

## Possible Solution

* Block everything, until the server is started. Maybe visualize with a better message that the language server is currently initializing
* Or formatting could work and don't care about `elm-analyze`. Since formatting changes the file this could lead to issues, like currently.

When opening the editor:

* Configure format on save and press save a bunch of times.
* Or: trigger file format manually multiple times.

## Context

Noticed this behavior multiple times. Other language servers (e.g. Java) take also a lot of time on startup where they're unusable.

## Your Environment

* Version used: 1.2.2
* Editor used: Vim/coc
* Environment name and version (e.g. node.js 5.4): node 11.11.0
* Operating System and version: Ubuntu 18.04

andys8:
This will not entirely make the startup perfect, but it looks to me, like it could be improving it.

vokoscreen-2019-08-10_14-52-04

@Razzeee Can you have a look?

Closes #125

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -29,14 +29,12 @@ connection.onInitialize(
     const { Server } = await import("./server");
     server = new Server(connection, params, parser);
     await server.init();
+    // Not awaiting the async result here!
+    server.registerInitializedProviders();
 
     return server.capabilities;
   },
 );
-
-connection.onInitialized(async () => {
-  server.registerInitializedProviders();
-});
 
 // Listen on the connection
 connection.listen();
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@elm-tooling/elm-language-server",
-  "version": "1.2.1",
+  "version": "1.2.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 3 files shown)


src/index.ts
```diff
@@ -24,15 +24,13 @@ connection.onInitialize(async (params: InitializeParams) => {
 
   const { Server } = await import("./server");
   server = new Server(connection, params, parser);
-  await server.registerInitializeProviders();
+  // Not awaiting the async result here!
+  server.registerInitializedProviders();
```
andys8: This is the code that changes behavior.

Razzeee: It's likely not the cause. I would probably group up the Providers we're sitting up by type. Those that need the imports and the forest to work and those that do not. Then run the ones that do not before we run `initWorkspace`

Razzeee:
Can you check if https://github.com/elm-tooling/elm-language-server/tree/restructure-init improves things?

PR #127, opened 2019-08-11
Setting: `diagnosticsOnSaveOnly` (bool)

andys8:
If enabled, diagnostics are only triggered on save, and not on change.
The default value is `false`, which means diagnostics are updated on change and on save.

Relates to #67


Razzeee:
Please rebase, even if github says it's fine

Issue #133, opened 2019-08-14
Configuration option to disable elm-analyse (linting)


andys8:
It could be useful to let the user decide to disable `elm-analyse`. Currently it can be configured to run only on save. A better configuration could be:

`"elmAnalyseTrigger" : "change" | "save" | "never"`

## Other language servers

There is hlintOn in Haskell land (hie): https://github.com/haskell/haskell-ide-engine/blob/master/README.md#configuration

## Mentions

@BrianHicks

andys8:
Solved by #139

PR #139, opened 2019-08-17
elm analyse configuration ("change" | "save" | "never").

andys8:
Adds configuration: `elmAnalyseTrigger` ("change" | "save" | "never"). Will not initialize `elm-analyse` if the setting is `never`. 

Solves #133

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,9 @@ This server contributes the following settings:
 - `elmLS.elmPath`: The path to your `elm` executable.
 - `elmLS.elmFormatPath`: The path to your `elm-format` executable.
 - `elmLS.elmTestPath`: The path to your `elm-test` executable.
-- `elmLS.diagnosticsOnSaveOnly`: Diagnostic updates triggered _only_ on save. `true/false` (default: `false`)
+- `elmLS.elmAnalyseTrigger`: `elm-analyse` executed on `'change'`, `'save'` or `'never'` (default: `'change'`)
+
+Settings may need a restart to be applied.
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -70,7 +70,7 @@ This server contributes the following settings:
 - `elmLS.elmPath`: The path to your `elm` executable.
 - `elmLS.elmFormatPath`: The path to your `elm-format` executable.
 - `elmLS.elmTestPath`: The path to your `elm-test` executable.
-- `elmLS.diagnosticsOnSaveOnly`: Diagnostic updates triggered _only_ on save. `true/false` (default: `false`)
+- `elmLS.elmAnalyseTrigger`: `elm-analyse` executed on `'change'`, `'save'` or `'never'` (default: `'change'`)
```
Razzeee: Mention, that it will likely need a server restart to take effect

Razzeee:
Needs a rebase, but tested fine for me.

README.md
andys8: Isn't this the case for any setting? We can add that line if you want to.

Razzeee: It depends, changing the paths should work without reload as the server tries to fetch fresh settings right before using them. Which doesn't apply to this, as it only happens in init.

andys8: Tested it by breaking "elmPath" and "elmFormatPath". They keep working without restart. Therefore the settings don't apply. So, I added a note.

andys8:
PR rebased and force-pushed.

PR #146, opened 2019-08-24
Rework  spawning of childprocesses

Razzeee:
This still needs some work. 
I'm aware of:
 - ~~It's not responding correctly when it can't find a bin with that name, instead it crashes~~

Needs testing:
- It will detect bins from global (as before). But it will also use bins from node dev dependencies, I only tested for these in the same folders as the elm.json. But hopefully it works for all parent dirs also (as is custom in npm land).
- ~~Can we still overwrite the settings correctly with an absolute path?~~ Works
- Needs testing on every platform

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ npm link
 
 ## Requirements
 
-You will need to install `elm` and `elm-test` to get all diagnostics and `elm-format` for formatting.
+You will need to install `elm` and `elm-test` to get all diagnostics and `elm-format` for formatting. Alternatively you can also just install these to your local npm `package.json`.
 
 ```sh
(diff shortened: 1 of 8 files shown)


Razzeee:
Pushed some improvements, only a few more tests needed.

Razzeee:
Pushed a change that assumes the default bin names, when settings are empty, so that we can disable `prefereLocal`, when the user actually wants a different path.

Razzeee:
Tested windows, now works fine with the latest fix.

I also tested if parent node installations will affect which version we end up with and it does, as expected. Great news.

Razzeee:
Got positive feedback from macOS, will merge soon.

PR #148, opened 2019-08-29
[Bugfix] execa stripping new lines resulted wrong elm-format diff

andys8:
* Changing the last line duplicated the last line
* Changing the first line duplicated the whole file

<https://github.com/sindresorhus/execa#stripfinalnewline>

vokoscreen-2019-08-29_21-04-06

### DIFF
--- a/src/util/elmUtils.ts
+++ b/src/util/elmUtils.ts
@@ -36,6 +36,7 @@ export async function execCmd(
       cwd,
       input,
       preferLocal,
+      stripFinalNewline: false,
     });
   } catch (error) {
     if (error.errno === "ENOENT") {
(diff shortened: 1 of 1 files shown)


Razzeee:
Can you make this configurable and only use it on elm make? As it's handy for the `elm --version` call

andys8:
Or explicit `trim()` in that case?

Razzeee:
Yes, if you want to. But please look at the git changes and just get it back from there, as it was like that before execa.

andys8:
Okay, thanks.

You're referring to 

```
  const version = result.stdout
    .substring(0, result.stdout.indexOf("\n"))
    .trim();
```

But I would think `.trim()` should be enough.

```

andys8:
Tested. Log says: `"message": "Elm version 0.19.0 detected."`

Razzeee:
Tested 
Manjaro with 0.19.0 and 0.19.1-beta-1
Windows with 0.19.0

Seems to work fine, no idea why I thought that substring was necessary.

PR #149, opened 2019-08-29
Readme: Explaining vim filetype

andys8:
It turned out to be necessary, that elm files are recognized as elm files. Most people install syntax highlighting anyway. But if not, elm-language-server wouldn't work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Just install the [`elm-tooling/elm-language-client-vscode`](https://github.com/e
 
 ### Vim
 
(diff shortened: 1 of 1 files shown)


Razzeee:
Thank you. On a related note do you think it's possible/worth it to package this all up?

andys8:
In general I'd prefer not to maintain "wrapper packages" like the one for ale. For vim in particular, everybody has a different config, and there is no way an aggregated plugin could be a good fit for multiple people.

But one idea could be to have a reference example configuration for elm only. It could also be used to test if the language server is working for vim.

Issue #147, opened 2019-08-29
Arguments: default to --stdio to reduce configuration


andys8:
Most config contain something like `"args": ["--stdio"]`

In order to achieve a tiny default configuration, we could default to this setting.

Compare e.g. Lua:

```
"languageserver": {
    "lua": {
      "command": "lua-lsp",
      "filetypes": ["lua"]
    }
  }
```

andys8:
Solved with #151

PR #151, opened 2019-08-29
Default argument `--stdio` if nothing is set

andys8:
To reduce configuration settings, it makes things easier, if there is a
default argument, if none is set. The result is that user configuration
can be simpler.

See vscode language server node implementation:

<https://github.com/microsoft/vscode-languageserver-node/blob/082631da44d6799358602a4abebadbf4fc47859b/server/src/main.ts#L1705>

Solves #147

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -11,6 +11,11 @@ import {
 import Parser from "web-tree-sitter";
 import { ILanguageServer } from "./server";
 
+// default setting `--stdio`
+if (process.argv.length === 2) {
+  process.argv.push("--stdio");
+}
+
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -11,6 +11,11 @@ import {
 import Parser from "web-tree-sitter";
 import { ILanguageServer } from "./server";
 
+// default setting `--stdio`
+if (process.argv.length === 2) {
+  process.argv.push("--stdio");
```
andys8: Looks both ugly and straight forward at the same time :D

PR #165, opened 2019-10-03
Print version with `-v` or `--version`

andys8:
`elm-language-server -v` or `--version` prints version.

### Side note

Normally `const { version } = require('../package.json')` can be used to retrieve the current version, but typescript will complain to import a file outside of its defined `rootDir`. That's why `project-version` is used.


Razzeee:
Looks fine.
Can we find a dependency that's a bit more updated and active?

andys8:
I think the dependency doesn't do much. If it's working, it should be fine.

I'd prefer to have no dependency for this, but I did some research and it didn't turn out to be as simple as I initially thought. More details in questions like this one: https://stackoverflow.com/questions/50822310/how-to-import-package-json-in-typescript

Razzeee:
Thank you

PR #171, opened 2019-10-15
Release instructions: Nix upstream

andys8:
Added nix, so it won't be forgotten and left behind

### DIFF
--- a/HOW_TO_RELEASE.md
+++ b/HOW_TO_RELEASE.md
@@ -5,3 +5,4 @@
 - Push the code to github and wait for the CI
 - Run `npm publish`
 - Tag the new version on github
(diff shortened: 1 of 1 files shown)


Razzeee:
Not sure if that "burden" should be placed here.

But basically we need to checkout the repo run https://github.com/turboMaCk/nixpkgs/blob/98997bb48997b27287a2995460d2fb6e1db88de7/pkgs/development/compilers/elm/packages/generate-node-packages.sh and push that up again?

andys8:
That should be how it's done.

I saw this ["Upgrade downstream projects"](https://scalameta.org/metals/docs/contributors/releasing.html) in the `scala/metals` project and thought it was a good idea. We can also add "Make sure that..." in front of it.

Razzeee:
We can try, I guess I'll get back to you after the next release :P

andys8:
We'll figure it out ^^

Issue #178, opened 2019-10-20
`--version` is not working as expected if installed via npm


andys8:
## Issue

It's using the `package.json` in the working directory...

### Directory without `package.json`
```shell
elm-language-server -v
/home/user/.npm-global/lib/node_modules/@elm-tooling/elm-language-server/node_modules/web-tree-sitte
var Module=void 0!==Module?Module:{};!function(e,t){"function"==typeof define&&define.amd?define([],t):
ule.hasOwnProperty(e)&&(t[e]=Module[e]);Module.arguments=[],Module.thisProgram="./this.program",Module.
a="function"==typeof importScripts,i="object"==typeof process&&"function"==typeof require&&!r&&!a,n=!r&
,l||(l=require("path")),e=l.normalize(e),n=o.readFileSync(e),t?n:n.toString()},Module.readBinary=functi
gv.length>1&&(Module.thisProgram=process.argv[1]

Error: ENOENT: no such file or directory, open '/home/user/dotfiles/package.json'
    at Object.openSync (fs.js:447:3)
    at Object.readFileSync (fs.js:349:35)
    at Function.module.exports.sync (/home/user/.npm-global/lib/node_modules/@elm-tooling/elm-langua
    at getVersion (/home/user/.npm-global/lib/node_modules/@elm-tooling/elm-language-server/node_mod
    at Object.<anonymous> (/home/user/.npm-global/lib/node_modules/@elm-tooling/elm-language-server/
    at Module._compile (internal/modules/cjs/loader.js:776:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:787:10)
    at Module.load (internal/modules/cjs/loader.js:643:32)
    at Function.Module._load (internal/modules/cjs/loader.js:556:12)
    at Module.require (internal/modules/cjs/loader.js:683:19)))))
```

## Directory with `package.json`

It shows the local version...

```
elm-language-server -v
1.0.0
```

## Related code

https://github.com/elm-tooling/elm-language-server/blob/5a39e7159ecb710e13f2d04102dbbeb27d2b289f/src/index.ts#L18

## Fixes

* Make the typescript import (disallowed by current config) of `package.json` work and replace dependency
* Copy current version into code or file in compile step
* Look for other dependencies that actually work as expected

BrianHicks:
I get this in the 1.4.2 built by nix, too.

Razzeee:
Doesn't seem to work on windows at all
[...]

PR #185, opened 2019-10-21
Bugfix: -v to print version was not working

andys8:
@razzee proposed to use `pjson` and it'll indeed work fine from other
directories

<https://github.com/elm-tooling/elm-language-server/issues/178#issuecomment-544523586>

Solves #178

Please verify @Razzeee  :)

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -24,3 +24,7 @@ jobs:
           npm test
         env:
           CI: true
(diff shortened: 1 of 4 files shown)


andys8:
Added the execution to github actions: Added the printing the version to github actions: https://github.com/elm-tooling/elm-language-server/pull/185/checks?check_run_id=268597308#step:5:10

PR #198, opened 2019-11-20
Improve import completions

Razzeee:


src/providers/completionProvider.ts
```diff
@@ -48,14 +48,74 @@ export class CompletionProvider {
...
       const replaceRange = Range.create(
         Position.create(params.position.line, currentCharacter),
         params.position,
       );
+
+      currentCharacter--;
+      const previousWordEnd = currentCharacter;
+      while (
+        currentCharacter - 1 >= 0 &&
+        (targetLine[currentCharacter - 1] !== " " &&
```
Razzeee: do we have more chars for this list?

Razzeee: `{`

PR #205, opened 2020-01-06
Fix multiple elm workspaces

### LINKED ISSUE #187: Support sub-projects / elm-review
## Expected Behavior
When opening a file in a sub-project, imports can be resolved.

Use case: [elm-review](https://package.elm-lang.org/packages/jfmengels/elm-review/latest/). elm-review puts a sub-project in the `review/` folder, next to your regular `src/`.

## Current Behavior
When opening a file in a sub-project, imports can not be resolved.

> MODULE NOT FOUND - You are trying to import a `Review.Rule` module:
>
> ...

```
# Install Elm and elm-review:
❯ echo '{"private":true}' > package.json
❯ npm i elm elm-review

# Make an Elm project:
❯ npx elm init
# Answer "Y".

# Set up elm-review:
❯ npx elm-review init
# Answer "Y".

# Make a simple Elm src file:
❯ echo 'module Main exposing (..)
  x = 1' > src/Main.elm

# Install packages for elm-review:
❯ cd review
❯ npx elm make ReviewConfig.elm
Starting downloads...
Dependencies ready!
Success! Compiled 1 module.

# Open VSCode for the whole repo:
❯ cd ..
❯ code .

# In VSCode:
#
# 1. Open src/Main.elm. Notice how "everything works".
# 2. Open review/ReviewConfig.elm. Notice how the `Review.Rule` import cannot be found.

# Open VSCode inside review/ instead:
❯ cd review
❯ code .

# In VSCode:
#
# 1. Open ReviewConfig.elm. Now the `Review.Rule` import can be found!
```

## Context
elm-review is really cool. It would be very nice to be able to just do `code .` in the repo root and be able to work on source code, tests and review rules all at the same time.

## Your Environment
* Version used: 0.7.1
* Editor name and version:  VSCode 1.39.2
* Environment name and version: Node.js v12.8.0
* Operating System and version: macOS Mojave

sentience:
Register `Provider`s in server.ts so each one has access to all of the `ElmWorkspace`s, and can process individual events in the relevant context.

Will fix #187.


src/elmWorkspace.ts
```diff
@@ -249,7 +249,10 @@ export class ElmWorkspace {
 
     return globby
       .sync(`${globUri}/**/*.elm`, { suppressErrors: true })
-      .map(matchingPath => ({ path: matchingPath, writable: element[1] }));
+      .map(matchingPath => ({
+        path: matchingPath,
+        writable: element[1],
+      }));
```
sentience: No change here. This is just Prettier formatting.

Razzeee:
Looks promising, as far as I can tell.

sentience:
This is now hypothetically feature-complete.

There is still a significant amount of refactoring to be done to tidy things up before this is merged, and considerable testing will be necessary to ensure anything that I've broken is fixed up. That said, anyone who likes to live on the edge, or who is particularly dependent on support for multiple elm.json files may wish to take this for a test drive.

In other words: bug reports welcome!

sentience:
OK, pretty happy with the abstractions now, although I'm still tossing up whether `Provider` should be a base class, or a separate object that is used to compose providers (`WorkspaceEventRouter`?).

My next step will be to fix two issues I've already observed in my testing (clinking the links in code lenses to show references to a function doesn't work, and some of my compiler errors are generating invalid file paths, preventing me from opening the relevant file by clicking the item in the Problems panel).

Once those are done, I'll be asking @Razzeee for advice on how to get this properly tested and released.

Razzeee:
I'll try to find to to review this in the next few days

Am Mo., 13. Jan. 2020 um 14:17 Uhr schrieb Kevin Yank <
notifications@github.com>:

sentience:
One of the two bugs I was experiencing turned out to be a separate issue with the language server. I've submitted a fix as #206.

Razzeee:
Hrm, clicking `x references` in a codelens seems to work fine for me. Is there something else that might affect your case? Do you get it too with the spa example?

sentience:
Interesting! I’m getting it with even a trivially-sized project (like https://github.com/sentience/elm-language-server-bug-demo).

I’m sure I’ll find the source of the problem; will just take me a little time to debug.

Razzeee:
What I'm seeing that does not work is having references across project borders. Not sure if we will be able to fix that ad-hoc.

sentience:
@Razzeee Can you share a representative project? Happy to have a crack at fixing it.

Razzeee:
I did test that on your https://github.com/sentience/elm-language-server-bug-demo repo, but I fixed the library function :)

sentience:
@Razzeee I spent hours tracing the issue I was having with Code Lenses, but after updating my working copy of elm-language-client-vscode to the latest master, I'm no longer able to reproduce it!

For the record (in case you might see what I was doing wrong), the issue I was having was that when I clicked a “n references” code lens, VS Code would display an error notification:

> argument does not match one of these constraints: arg instanceof constraint, arg.constructor === constraint, nor constraint(arg) === true

I traced this to [a line in VS Code](https://github.com/microsoft/vscode/blob/4b1fdfbdbf47703e749fb44c70bbaa6eb6306114/src/vs/base/common/types.ts#L198), which gets called to check that the types of the arguments provided to a `Command` are the types that are required by that type of `Command`.

The command in question is this one: https://github.com/elm-tooling/elm-language-server/blob/ff1b5124405d943f0f870beddbac0006e1e815ee/src/providers/codeLensProvider.ts#L118-L128

According to the [VS Code documentation](https://code.visualstudio.com/api/references/commands), `editor.action.showReferences` takes three arguments:

> * _uri_ - The text document in which to show references
> * _position_ - The position at which to show
> * _locations_ - An array of locations.

…but in the code above, the command gets created with a single argument: an object with three fields `range`, `references`, and `uri`. The error I was encountering was VS Code receiving this object as the first (and only) argument to the command and checking if it was a `URI` (as the first argument should be).

Why this error was happening for my code (which had made no change to this command) and not yours was a mystery to me. Nevertheless, I began the process of converting the command's arguments to the appropriate format. This proved surprisingly hard, as I discovered annoying differences in things like line numbers (which VS Code was expecting to be 1-based, but which were 0-based in the types used by the Language Server Protocol).

Finally I chanced upon https://github.com/Microsoft/vscode/issues/67649, which had been posted by someone equally confused as me about the type differences between VS Code's internals and the Language Server Protocol. The [advice there](https://github.com/Microsoft/vscode/issues/67649#issuecomment-459365333) was to stick to the LSP data formats, and to rely on some [“automagical” conversion code buried deep inside VS Code](https://github.com/Microsoft/vscode/blob/70160d0e580d0588e37c93fe8724014d76b00939/src/vs/workbench/api/node/extHostCommands.ts#L110-L138).

For whatever reason, those type conversions were apparently not working for me in my working copy. In a moment of desperation, I pulled the latest version of master for elm-language-client-vscode, and found that the issue was resolved! I have no idea what could have changed to make a difference. I do notice you bumped the version of TypeScript for the project to 3.7.4, but I had already upgraded to 3.7.3 in my working copy, so I was on almost the exact same version.

Any ideas? :)

In any case, I've now resolved all the issues I'm seeing in my testing. I just need to take a look at [the cross-project reference issue](https://github.com/elm-tooling/elm-language-server/pull/205#issuecomment-577578841) that you mentioned now.

sentience:
@Razzeee This is a screenshot of a reference from Main.elm to Library.elm in my sample project (with the library function fixed) properly showing the definition of `Library.library` in a pop-over. ⌥-clicking the function name opens the Library.elm file at the right location.

Can you give me some more detail on the issue you've found?

_Extension_Development_Host__-_src_Main_elm_—_elm-language-server-bug-demo

Razzeee:
Never mind, just tested again. Reference counting works fine. I guess it hung due to some reason.

Let's get this merged! :)

sentience:
OK, I’ll tidy up the branch history and then it should be ready to go!

sentience:
@Razzeee OK, this should be good to go. I did a little more refactoring (moving the shared logic between providers from a `Provider` superclass to a composed `ElmWorkspaceMatcher` object). Apart from that, no code changes here since you reviewed it.

I've tidied up the commit history and included a package version bump and changelog for a 1.5.0 release. I'm happy for you to merge this now if you are!

Razzeee:
I'm planning to review and merge this tomorrow.

src/providers/noWorkspaceContainsError.ts
```diff
@@ -0,0 +1,7 @@
+import { URI } from "vscode-uri";
+
+export class NoWorkspaceContainsError extends Error {
```
Razzeee: I can only find code checking for this, but are these created somewhere or was it changed and is now obsolete? :)

Razzeee:
Thanks for this and all the drive by improvements, your awesome :unicorn: 

I'll merge this now, we can follow up the one thing I remarked (might just be my understanding)

src/providers/noWorkspaceContainsError.ts
sentience: Good catch! Fixed by #208.

PR #210, opened 2020-01-31
Update packages ranking

rlefevre:
Ranking algorithm has been improved and uses more than 4300 OSS applications.


Razzeee:
Thank you @rlefevre

PR #214, opened 2020-02-03
Improve loading progress

sentience:
Fixes several problems with displaying loading progress, especially across multiple elm.json files.

Also resolves issue where the language server was attempting to process events before it had finished initialising, causing errors and failure to load features in files open during initialisation.

Caveat: In large projects there still seems to be a stall in loading at the point where a certain number of elm-tree-sitter processes have been spawned. This is not a new problem, but the visible loading progress makes it more visible, as the load percentage stalls. Would like to look into what's going on there in a later piece of work.

### DIFF
--- a/src/elmWorkspace.ts
+++ b/src/elmWorkspace.ts
@@ -184,21 +184,24 @@ export class ElmWorkspace {
       }
 
       const promiseList: Promise<void>[] = [];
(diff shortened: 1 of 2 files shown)


Razzeee:
Awesome, thank you.

Some embarrassing oversights on my end :)

PR #220, opened 2020-03-17
Add If-Else snippet

### LINKED ISSUE #219: Add if else snippet and completion support
Currently there is no if then else support, which makes it difficult to write them. A snippet would be nice. Also, the word `then` always autocompletes to `Maybe.andThen` when pressing enter or tab, which makes it hard to write if else statements.

jmbockhorst:
Part of #219.


Razzeee:
Awesome, thank you! 🌈

PR #221, opened 2020-03-17
Add record completion items

### LINKED ISSUE #218: Record field values should have completion items when updating
When updating a record, the field values should have completion items.

elm-record

In this case, the top choice should be `lastName`.

jmbockhorst:
Fixes #218. Adds record completion items for that type when updating a record. Also works for imported records.

record-completion-items

### DIFF
--- a/src/providers/completionProvider.ts
+++ b/src/providers/completionProvider.ts
@@ -450,12 +450,21 @@ export class CompletionProvider {
     uri: string,
   ): CompletionItem[] {
     const result: CompletionItem[] = [];
-    const typeDeclarationNode = TreeUtils.getTypeAliasOfRecord(
+    let typeDeclarationNode = TreeUtils.getTypeAliasOfRecord(
       node,
       tree,
       imports,
       uri
     );
+
+    if (!typeDeclarationNode && node.parent?.parent) {
+      typeDeclarationNode = TreeUtils.getTypeAliasOfRecordField(
+        node.parent.parent,
+        tree,
+        imports,
+        uri
+      );
+    }
 
     if (typeDeclarationNode) {
       const fields = TreeUtils.getAllFieldsFromTypeAlias(
--- a/src/providers/completionProvider.ts
+++ b/src/providers/completionProvider.ts
@@ -162,7 +162,7 @@ export class CompletionProvider {
       } else if (
         nodeAtPosition.parent?.parent?.type == "record_expr"
       ) {
-        return this.getRecordCompetions(nodeAtPosition, tree, replaceRange, elmWorkspace.getImports(), params.textDocument.uri);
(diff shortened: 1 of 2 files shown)


src/providers/completionProvider.ts
```diff
@@ -159,6 +159,10 @@ export class CompletionProvider {
           nodeAtPosition.parent.parent,
           replaceRange,
         );
+      } else if (
+        nodeAtPosition.parent?.parent?.type == "record_expr"
+      ) {
+        return this.getRecordCompetions(nodeAtPosition, tree, replaceRange, elmWorkspace.getImports(), params.textDocument.uri);
```
Razzeee: Typo in `Completion`

src/util/treeUtils.ts
```diff
@@ -1159,6 +1159,75 @@ export class TreeUtils {
...
+      node &&
+      node.previousNamedSibling?.type === "type_annotation"
+    ) {
+      const typeAnnotationNodes = TreeUtils.findAllNamedChildrenOfType(
+        ["type_ref", "type_expression"],
+        node.previousNamedSibling
+      );
+      if (typeAnnotationNodes) {
+        const type = typeAnnotationNodes[typeAnnotationNodes.length - 1];
+        return type.firstNamedChild?.firstNamedChild || type;
```
Razzeee: While it might be unproblematic in this case, I think using `??` instead of `||` would be nicer

src/providers/completionProvider.ts
```diff
@@ -438,6 +442,50 @@ export class CompletionProvider {
...
+      )?.text || "";
+
+      fields?.forEach(element => {
+        const hint = HintHelper.createHintForTypeAliasReference(
+          element.type,
+          element.field,
+          typeName,
+        );
+        result.push(
+          this.createFunctionParameterCompletion(
```
Razzeee: I think we either want our own `create` method for record fields or we want to rename the `functionParameter` one to be a bit broader, what do you think?

Razzeee:
Awesome work, there are some very minor things, we should address before merging.

jmbockhorst:
Thanks for the feedback. This is my first contribution to a language server, so any guidance is appreciated. I made the requested changes.

src/providers/completionProvider.ts
jmbockhorst: I went with `createFieldOrParameterCompletion` since I think the method covers both cases. Let me know if you think there is a better name.

jmbockhorst:
I also added support for nested record assignment completions. Like this:
```
{ model |
     data1 = {
            field (Show completions for fields for the type data1)
     }

}

```

Razzeee:
Looks good to me, I'll probably merge this later and run some tests in master then :)

PR #222, opened 2020-03-18
Add basic keyword completions

### LINKED ISSUE #219: Add if else snippet and completion support

jmbockhorst:
Fixes #219. I wasn't sure about the sort order for keywords. I set the prefix to `a`, since it seems like in TypeScript the keywords are mostly towards the top, but it might not be correct.

### DIFF
--- a/src/providers/completionProvider.ts
+++ b/src/providers/completionProvider.ts
@@ -177,6 +177,7 @@ export class CompletionProvider {
       );
 
       completions.push(...this.createSnippets());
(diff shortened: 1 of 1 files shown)


Razzeee:
Perfect, we can still improve the sorting later on

PR #223, opened 2020-03-18
Fix hovers for custom types

### LINKED ISSUE #209: Type info in popup
Show type info popup inside `case .. of` or `function`

## Expected Behavior
Atom:
Inside `case .. of`
image

image

image

Inside a `function`
image

## Current Behavior
No popup in `case/function` for custom types.

## Context
Related to https://github.com/elm-tooling/elm-language-server/issues/202

## Your Environment
VSCode with Elm 0.8.0 plugin

All these cases look similar to me, but if you want these in separate issues let me know.

jmbockhorst:
I think this fixes #209. Shows hovers for custom type aliases and types (when there isn't a block comment above it). I wasn't thrilled with the hovers for the union variants because of the lack of syntax coloring, but I think it is better than nothing.

### DIFF
--- a/src/util/hintHelper.ts
+++ b/src/util/hintHelper.ts
@@ -60,11 +60,22 @@ export class HintHelper {
       let code: string | undefined;
       let comment: string = "";
       let annotation: string = "";
-      if (declaration.type === "type_declaration") {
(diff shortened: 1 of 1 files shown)


Razzeee:
Yeah, we would need to change the syntax highlighting to pick that up too, but might be a bit problematic. Everything else looks fine.

PR #225, opened 2020-03-19
Handle onPrepareRename to only allow renaming when it can be done

### LINKED ISSUE #180: Disable rename for things that cannot be renamed
## Expected Behavior
If I press <kbd>F2</kbd> on something that does not support rename, "You cannot rename this element" should be shown, like in TypeScript.

## Current Behavior
As long as there is a word under the cursor, <kbd>F2</kbd> initiates a rename. After typing in a new name and pressing <kbd>Enter</kbd>, "No result." is shown.

Starting code:

```elm
module Main exposing (..)


foo m =
    case m of
        Just string ->
            string

        Nothing ->
            "nothing"

```

1. Put the cursor on either of the two occurances of the word `string`.
2. Press <kbd>F2</kbd> to start a rename.
3. Enter `str` and press <kbd>Enter</kbd>.

"No result" is shown. Until renaming for "pattern variables" is implemented, it’d be nicer if "You cannot rename this element" was shown immediately when pressing <kbd>F2</kbd>.

Similarly, try "renaming" `case` or the word `nothing` in the string. It doesn’t make sense to rename those! TypeScript shows "You cannot rename this element" if you try to rename a keyword or something inside a string.

## Context
Renaming "pattern variables" is something I’ve tried to do several times and have been disappointed when I only get "No result" after having typed in the new name. Same thing with record fields.

Renaming keywords (such as `case`) and words inside strings is nothing I’ve done in real life – it’s just something I thought of when writing this issue :)

## Your Environment

- Version used: 0.6.0
- Editor name and version: VSCode 1.39.2
- Environment name and version: Node.js v12.8.0
- Operating System and version: macOS Mojave

jmbockhorst:
Fixes #180.

### DIFF
--- a/src/capabilityCalculator.ts
+++ b/src/capabilityCalculator.ts
@@ -38,7 +38,9 @@ export class CapabilityCalculator {
       foldingRangeProvider: true,
       hoverProvider: true,
       referencesProvider: true,
-      renameProvider: true,
+      renameProvider: {
+        prepareProvider: true
+      },
       selectionRangeProvider: true,
       textDocumentSync: TextDocumentSyncKind.Full,
       workspaceSymbolProvider: true,
--- a/src/providers/renameProvider.ts
+++ b/src/providers/renameProvider.ts
@@ -5,9 +5,10 @@ import {
   RenameParams,
   TextEdit,
   WorkspaceEdit,
+  PrepareRenameParams,
 } from "vscode-languageserver";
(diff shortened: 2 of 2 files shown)


Razzeee:
Seems to work great, thank you. I'm probably going to extend this to give a reason in case you try to edit stuff outside of your project folder (from dependencies etc)

Issue #194, opened 2019-11-08
Feature: Move function/type


jonathanmoregard:
## Expected Behavior
When activating quickfix through the vscode command "editor.action.quickFix" while the cursor is on a function declaration or on a  type declaration; I want the option to move that function/type to another file, automatically updating the import statements on all files where the function/type is used.

## Current Behavior
No move action available on "quickFix"

## Context
When refractoring big elm apps, sometimes I need to move functions between files, to keep a clear separation of concerns.

Right now, this is a lot of manual work, where I:
1. cut the functiuon/type from module 1
2. paste the function/type in module 2
3. expose the function/type in module 2
4. unexpose the function/type in module 1
3. compile
4. go through all import-errors one by one, adjusting the imports to point to module 2 instead of module 1

Razzeee:
Functions can be moved since march, types still need to be implemented

razzeee:
LSP 3.18 now defines `CodeActionKind.RefactorMove`. The existing Move Function action is implemented but currently reports `refactor.rewrite` and depends on the custom `moveFunctionRefactoringSupport` capability.

As part of this issue, we should:

- classify Move Function, and the future Move Type action, as `refactor.move`;
- advertise/document the standard kind so clients can filter it correctly;
- use `WorkspaceEditMetadata.isRefactoring` where supported;
- retain the custom destination-selection request only for the UI that the standard protocol does not provide.

PR #226, opened 2020-03-20
Add basic support for move function refactoring

jmbockhorst:
Part of #194. It adds basic support for a new code action to move a function to a new file. It is a fair amount of changes so I wanted to get your feedback to make sure it was a good direction. 

Because this type of refactoring requires client support (quick pick for the destination file),
I added a new `extendedCapabilites` property to the client settings that the client can initialize. Right now the only property is `moveFunctionRefactoringSupport`, but more can easily be added.

I also added a `protocol.ts` that defines the custom requests between the server and client. 

Currently, in vscode, you can move a function to a new file and it adds new imports to all files with references. Dependent on https://github.com/elm-tooling/elm-language-client-vscode/pull/88.

### DIFF
--- a/src/providers/handlers/moveRefactoringHandler.ts
+++ b/src/providers/handlers/moveRefactoringHandler.ts
@@ -1,169 +1,179 @@
-import { IConnection, TextEdit, Range, Position } from 'vscode-languageserver';
-import { GetMoveDestinationRequest, MoveDestinationsResponse, MoveParams, MoveRequest, MoveDestination } from '../../protocol';
-import { ElmWorkspace } from '../../elmWorkspace';
-import { URI } from 'vscode-uri';
-import { ElmWorkspaceMatcher } from '../../util/elmWorkspaceMatcher';
-import { TreeUtils } from '../../util/treeUtils';
-import { References } from '../../util/references';
+import { IConnection, TextEdit, Range, Position } from "vscode-languageserver";
+import {
+  GetMoveDestinationRequest,
+  MoveDestinationsResponse,
+  MoveParams,
+  MoveRequest,
+  MoveDestination,
+} from "../../protocol";
+import { ElmWorkspace } from "../../elmWorkspace";
+import { URI } from "vscode-uri";
+import { ElmWorkspaceMatcher } from "../../util/elmWorkspaceMatcher";
+import { TreeUtils } from "../../util/treeUtils";
+import { References } from "../../util/references";
 
 export class MoveRefactoringHandler {
-	constructor(
-		private connection: IConnection,
-		private elmWorkspaces: ElmWorkspace[]
-	) {
-		this.connection.onRequest(GetMoveDestinationRequest,
-			new ElmWorkspaceMatcher(elmWorkspaces, (param: MoveParams) =>
-				URI.parse(param.sourceUri),
-			).handlerForWorkspace(this.handleGetMoveDestinationsRequest.bind(this)));
+  constructor(
+    private connection: IConnection,
+    private elmWorkspaces: ElmWorkspace[],
+  ) {
+    this.connection.onRequest(
+      GetMoveDestinationRequest,
+      new ElmWorkspaceMatcher(elmWorkspaces, (param: MoveParams) =>
+        URI.parse(param.sourceUri),
+      ).handlerForWorkspace(this.handleGetMoveDestinationsRequest.bind(this)),
+    );
 
-		this.connection.onRequest(MoveRequest,
-			new ElmWorkspaceMatcher(elmWorkspaces, (param: MoveParams) =>
-				URI.parse(param.sourceUri),
-			).handlerForWorkspace(this.handleMoveRequest.bind(this)));
-	}
+    this.connection.onRequest(
+      MoveRequest,
+      new ElmWorkspaceMatcher(elmWorkspaces, (param: MoveParams) =>
+        URI.parse(param.sourceUri),
+      ).handlerForWorkspace(this.handleMoveRequest.bind(this)),
+    );
+  }
 
-	private handleGetMoveDestinationsRequest(
-		params: MoveParams,
-		elmWorkspace: ElmWorkspace
-	): MoveDestinationsResponse {
-		const forest = elmWorkspace.getForest();
+  private handleGetMoveDestinationsRequest(
+    params: MoveParams,
+    elmWorkspace: ElmWorkspace,
+  ): MoveDestinationsResponse {
+    const forest = elmWorkspace.getForest();
 
-		const destinations: MoveDestination[] = forest.treeIndex
-			.filter(tree => tree.writeable && tree.uri !== params.sourceUri)
-			.map(tree => {
-				let uri = URI.parse(tree.uri).fsPath;
-				const rootPath = elmWorkspace.getRootPath().fsPath
+    const destinations: MoveDestination[] = forest.treeIndex
+      .filter(tree => tree.writeable && tree.uri !== params.sourceUri)
+      .map(tree => {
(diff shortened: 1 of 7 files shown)


Razzeee:
Approach looks okay to me, it's unfortunate, that it needs client support, but that's how it is with language server at the moment. Until this moves further https://github.com/microsoft/language-server-protocol/issues/764

Can you please install and use prettier via https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode 
You seem to be using tabs and some other styling that is inconsistent to our prettier rules.

src/providers/handlers/moveRefactoringHandler.ts
```diff
@@ -0,0 +1,169 @@
...
+					const refTree = forest.getTree(refUri);
+
+					if (refTree) {
+						const importNodes = TreeUtils.findAllNamedChildrenOfType(
+							"import_clause",
+							refTree.rootNode
+						);
+
+						if (importNodes?.length) {
+							const lastImportNode = importNodes[importNodes?.length - 1];
```
Razzeee: Please create a helper in treeUtils from this. We will need this function for other stuff :)

jmbockhorst:
The core functionality is complete. Moving a function properly updates the module `exposing` values for the old and new files, and removes and adds the proper imports for all references. I tested this for a while and it seemed to work pretty well, but there is sure to be an edge case that could break it.

Razzeee:
1. I'm not convinced, auto exposing it in every file is the way to go, especially if it was never used like this. 

2. If it has been fully referenced e.g. `Feed.viewPagination` doesn't get changed to `xxx.viewPagination`.

3. Doc comments doesn't seem to be moved together with the function.

I think we should at least fix 1 and 3 before we merge this.

jmbockhorst:
What do you mean by 1? Auto exposing the moved function in the destination file? Currently it only gets exposed if there is an external reference that needs it.

Razzeee:
Kinda correct, it correlates with 2, as you don't seem to differentiate between fully qualified (`Feed.viewPagination`) and exposed (`viewPagination`)

I tested it with the elm-spa-example https://github.com/rtfeldman/elm-spa-example/blob/master/src/Article/Feed.elm#L172

jmbockhorst:
We also need to update references or add imports for any functions or types used within the moved function.

Razzeee:
Yes, would be nice, but I think it would already be useful without, as Elm will just complain to you about it :)

jmbockhorst:
So to be clear, if a file only has fully qualified references to the function (`xxx.function`), we don't need to expose it (just import). Only expose it if we need it, when there is an exposed reference to it in the file (`function`).

Razzeee:
Yes, that's my understanding

jmbockhorst:
It now properly handles fully qualified references:
- Don't expose the import if all uses are fully qualified
- Change the module name of all fully qualified references

src/providers/handlers/moveRefactoringHandler.ts
```diff
@@ -216,19 +222,49 @@ export class MoveRefactoringHandler {
...
+            .some(ref => !ref.fullyQualified);
+
           const refTree = forest.getTree(refUri);
 
           if (refTree) {
             const importEdit = RefactorEditUtils.addImport(
               refTree,
               destinationModuleName,
-              functionName,
+              needToExpose ? functionName : "",
```
Razzeee: I think you want the else to be `undefined`, due to your `addImport` implementation taking `undefined` as third param?

src/util/refactorEditUtils.ts
```diff
@@ -68,15 +68,71 @@ export class RefactorEditUtils {
...
-        `import ${moduleName} exposing (${valueName})`,
+        valueName
+          ? `import ${moduleName} exposing (${valueName})`
+          : `import ${moduleName}`,
+      );
+    }
+  }
+
+  public static changeQualifiedReferenceModule(
+    node: SyntaxNode,
```
Razzeee: `importNode` might make this easier to read?

src/util/refactorEditUtils.ts
```diff
@@ -68,15 +68,71 @@ export class RefactorEditUtils {
...
+    moduleName: string,
```
Razzeee: What do you think about `newModuleName`?

src/providers/handlers/moveRefactoringHandler.ts
jmbockhorst: Yep, that would be better.

src/util/refactorEditUtils.ts
```diff
@@ -68,15 +68,71 @@ export class RefactorEditUtils {
...
+            ),
+          ),
+          moduleName,
+        );
+      }
+    }
+  }
+
+  public static removeQualifiedReference(
+    node: SyntaxNode,
```
Razzeee: `importNode`?

src/util/refactorEditUtils.ts
jmbockhorst: Yeah, I think I had that at first but removed it for simplicity. I will change it back.

src/util/refactorEditUtils.ts
Razzeee: Nevermind, I realize now, that we're talking modules, not imports here

src/util/refactorEditUtils.ts
```diff
@@ -68,15 +68,71 @@ export class RefactorEditUtils {
...
+              moduleNode.endPosition.column,
+            ),
+          ),
+          moduleName,
+        );
+      }
+    }
+  }
+
+  public static removeQualifiedReference(
```
Razzeee: Haven't tested this, but might this lead to empty brackets being left behind? Like `import Html exposing ()`

src/util/refactorEditUtils.ts
jmbockhorst: In this context, node is actually just the `func` node of the `Module.func`.

src/util/refactorEditUtils.ts
jmbockhorst: This function just removes the module name from `Module.func`, so that it is just `func`.

src/util/refactorEditUtils.ts
Razzeee: It's hard for me to tell the context, will do a proper review and test later this evening on.

Razzeee:
So I ended up with this as a module declaration :)
`module Session exposing (Session, changes, cred, fromViewer, navKey, viewer), viewPagination
`

Razzeee:
> So I ended up with this as a module declaration :)
> `module Session exposing (Session, changes, cred, fromViewer, navKey, viewer), viewPagination `

Not able to do that again, but every time I use the declaration (not the type declaration) to initalize the move on viewPagination to the Session module I end up with `module Session exposing (Session, changes, cred, fromViewer, navKey, viewer, viewPagination toMsg page (Model feed))`

jmbockhorst:
Thanks, I think I fixed it.

src/util/treeUtils.ts
```diff
@@ -1181,14 +1188,18 @@ export class TreeUtils {
     node: SyntaxNode | undefined,
     tree: Tree,
     imports: IImports,
-    uri: string
+    uri: string,
   ): SyntaxNode | undefined {
-    if (node?.parent?.type == "function_call_expr" && node.parent.firstNamedChild) {
-      const parameterIndex = node.parent.namedChildren.map(c => c.text).indexOf(node.text) - 1;
+    if (
+      node?.parent?.type == "function_call_expr" &&
```
Razzeee: `===` please

Razzeee:
Merging this now, will address some of the smaller stuff in one of the next commits.

Thank you so much! 🌈

PR #229, opened 2020-03-24
Add the ability to expose/unexpose functions and types

### LINKED ISSUE #192: Feature: Expose/Unexpose functions
## Expected Behavior
When activating quickfix through the vscode command "editor.action.quickFix" while the cursor is on a function;

If the function is **exposed**, I want a context menu with the option of **unexposing** it.
If the function is **not exposed**, I want a context menu with the option of **exposing** it.

## Current Behavior
No expose/unexpose action available through quickfix.

## Context
When refractoring code, I spend a lot of time going to the top of files to expose/unexpose functions I'm working with, and my scroll finger gets sad and tired.

jmbockhorst:
Adds code actions and code lenses for exposing and unexposing functions and types. Closes #192. 

Code lenses need client side support since they can only send commands, not workspace edits. (https://github.com/elm-tooling/elm-language-client-vscode/pull/91)

### DIFF
--- a/src/providers/codeActionProvider.ts
+++ b/src/providers/codeActionProvider.ts
@@ -122,10 +122,6 @@ export class CodeActionProvider {
         if (edit) {
           codeActions.push({
             title: "Unexpose Function",
-            command: Command.create("Unexpose", "elm.unexpose", {
-              uri: params.textDocument.uri,
-              name: functionName,
-            }),
             edit: {
               changes: {
                 [params.textDocument.uri]: [edit],
--- a/src/providers/codeActionProvider.ts
+++ b/src/providers/codeActionProvider.ts
@@ -4,9 +4,10 @@ import {
   CodeActionParams,
   ExecuteCommandParams,
   IConnection,
+  Command,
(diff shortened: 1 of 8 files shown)


src/providers/codeActionProvider.ts
```diff
@@ -36,6 +40,9 @@ export class CodeActionProvider {
       // tslint:disable-next-line: no-unused-expression
       new MoveRefactoringHandler(this.connection, this.elmWorkspaces);
```
Razzeee: Doing this twice now, seems like an oversight

jmbockhorst: Whoops, that must've got left in from a merge conflict.

Razzeee:
A case where this fails is, when I try to unexpose something like `Page(..)`

jmbockhorst:
I wasn't sure if the default behavior for union types should be to expose `Type(..)` rather than just `Type` (current behavior).

Razzeee:
I also noticed that and am unsure too. I think it's fine, you will have to
edit it for some cases anyway.

PR #230, opened 2020-03-25
Add auto import code actions

### LINKED ISSUE #105: [Feature] Add missing imports as fix
Whenever there is a `Naming Error`:

```
[Elm] [E] NAMING ERROR - I cannot find a `Color` type:

13|     { white : Color
                  #^^^^^#
These names seem close though:

    #Bool#
    #Char#
    #Cmd#
    #Float#

#Hint#: Read <https://elm-lang.org/0.19.0/imports> to see how `import`
declarations work in Elm.
```

It can be a typo, but it's also possible something wasn't imported. The language server could match (in this case) `Color` against what is known in source files and dependencies.

If there is a hit, it could be suggested as a fix.

If the fix is applied, it can add a new line `import Module exposing (Color)` as second line and trigger `elm-format` which will do the merging of imports.

In my opinion this is one of the missing killer features :)

### LINKED ISSUE #193: Feature: auto import
## Expected Behavior
When activating quickfix through the vscode command "editor.action.quickFix" while the cursor is on a call site for an **unimported** function or type; I want the option to import the module containing the function/type.

The Haskell Language Server for haskell (https://marketplace.visualstudio.com/items?itemName=alanz.vscode-hie-server) does this pretty well  

Like this (Maybe is not imported here, quickFix allows me to import it, so that mapMaybe is available):
image
(command for displaying context menu: "editor.action.quickFix")

## Current Behavior
No import action available through quickfix.

## Context
When coding, I spend a lot of time importing things as new needs arise. If this could be done automatically I'd be a very happy person

jmbockhorst:
Adds code actions for auto importing. `Import 'value' from module "Module"` and ` Add all missing imports`. Closes #193, closes #105.

### DIFF
--- a/src/providers/diagnostics/elmMakeDiagnostics.ts
+++ b/src/providers/diagnostics/elmMakeDiagnostics.ts
@@ -17,6 +17,8 @@ import { Settings } from "../../util/settings";
 import { IElmIssue } from "./diagnosticsProvider";
 import { ElmDiagnosticsHelper } from "./elmDiagnosticsHelper";
 import execa = require("execa");
+import { RefactorEditUtils } from "../../util/refactorEditUtils";
+import { ImportUtils } from "../../util/importUtils";
 
 const ELM_MAKE = "Elm";
 const RANDOM_ID = randomBytes(16).toString("hex");
@@ -58,6 +60,10 @@ export interface IStyledString {
 
 export class ElmMakeDiagnostics {
   private elmWorkspaceMatcher: ElmWorkspaceMatcher<URI>;
+  private neededImports: Map<
+    string,
+    { moduleName: string; valueName: string; diagnostic: Diagnostic }[]
+  > = new Map();
 
   constructor(
     private connection: IConnection,
@@ -76,7 +82,7 @@ export class ElmMakeDiagnostics {
     const workspaceRootPath = this.elmWorkspaceMatcher
       .getElmWorkspaceFor(filePath)
       .getRootPath();
-    return await this.checkForErrors(
+    const diagnostics = await this.checkForErrors(
       workspaceRootPath.fsPath,
       filePath.fsPath,
     ).then((issues) => {
@@ -84,6 +90,49 @@ export class ElmMakeDiagnostics {
         ? new Map([[filePath.toString(), []]])
         : ElmDiagnosticsHelper.issuesToDiagnosticMap(issues, workspaceRootPath);
     });
+
+    // Handle import all
+    const forest = this.elmWorkspaceMatcher
+      .getElmWorkspaceFor(filePath)
+      .getForest();
+
+    const exposedValues = ImportUtils.getPossibleImports(forest);
+
+    // Get all possible imports from the diagnostics for import all
+    diagnostics.forEach((diagnostics, uri) => {
+      const sourceTree = forest.getByUri(uri);
+      this.neededImports.set(uri, []);
+
+      diagnostics.forEach((diagnostic) => {
+        if (diagnostic.message.startsWith("NAMING ERROR")) {
+          const text = sourceTree?.tree.rootNode.descendantForPosition(
+            {
+              column: diagnostic.range.start.character,
+              row: diagnostic.range.start.line,
+            },
+            {
+              column: diagnostic.range.end.character,
+              row: diagnostic.range.end.line,
+            },
+          ).text;
+
+          // Find imports
+          exposedValues
+            .filter((exposed) => exposed.value === text)
+            .forEach((exposed, i) => {
+              if (i === 0) {
+                this.neededImports.get(uri)?.push({
+                  moduleName: exposed.module,
+                  valueName: exposed.value,
+                  diagnostic,
+                });
+              }
+            });
+        }
+      });
+    });
+
+    return diagnostics;
   };
 
   public onCodeAction(params: CodeActionParams): CodeAction[] {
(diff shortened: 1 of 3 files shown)


src/providers/diagnostics/elmMakeDiagnostics.ts
```diff
@@ -76,14 +82,57 @@ export class ElmMakeDiagnostics {
...
+    const forest = this.elmWorkspaceMatcher
+      .getElmWorkspaceFor(filePath)
+      .getForest();
+
+    const exposedValues = ImportUtils.getPossibleImports(forest);
+
+    // Get all possible imports from the diagnostics for import all
+    diagnostics.forEach((diagnostics, uri) => {
+      const sourceTree = forest.getByUri(uri);
+      this.neededImports.set(uri, []);
```
Razzeee: Wouldn't this cause a reset, if you have multiple diagnostics in one file?

src/providers/diagnostics/elmMakeDiagnostics.ts
```diff
@@ -76,14 +82,57 @@ export class ElmMakeDiagnostics {
...
+
+      diagnostics.forEach((diagnostic) => {
+        if (diagnostic.message.startsWith("NAMING ERROR")) {
+          const text = sourceTree?.tree.rootNode.descendantForPosition(
```
Razzeee: This might work fine, but my impulse would have been to reach for `TreeUtils.getNamedDescendantForPosition` as it's supposed to cover some edgecases (which you might not have in this usecase)

src/providers/diagnostics/elmMakeDiagnostics.ts
```diff
@@ -76,14 +82,57 @@ export class ElmMakeDiagnostics {
...
+            },
+            {
+              column: diagnostic.range.end.character,
+              row: diagnostic.range.end.line,
+            },
+          ).text;
+
+          // Find imports
+          exposedValues
+            .filter((exposed) => exposed.value === text)
```
Razzeee: As your only processing the first item you find, I think you don't really want `filter` but `find` instead?

src/providers/diagnostics/elmMakeDiagnostics.ts
Razzeee: Nevermind.

Razzeee:
It seems to add one import for every diagnostic to the imports

image

src/util/refactorEditUtils.ts
```diff
@@ -136,6 +136,27 @@ export class RefactorEditUtils {
...
+    tree: Tree,
+    importData: {
+      moduleName: string;
+      valueName: string;
+    }[],
+  ): TextEdit | undefined {
+    const lastImportNode = TreeUtils.getLastImportNode(tree);
+
+    const imports = importData
+      .map(data => `import ${data.moduleName} exposing (${data.valueName})`)
```
Razzeee: We could just filter here

src/providers/diagnostics/elmMakeDiagnostics.ts
jmbockhorst: It actually could be multiple values from different modules, and we want to add all of those.
For example:
`import 'Html' from module "Html"`
`import 'Html' from module "Html.Styled"`

jmbockhorst:
Thanks, I ended up having to filter both in the `addImports` function and when we are checking if the "Add all imports" code action is needed. This way it only shows "Add all imports" when there are two or more separate types of imports.

I can't filter once before everything because we don't which diagnostic will be the one getting the code action request and it messes up the sorting (They are sorted by the diagnostic message).

src/providers/diagnostics/elmMakeDiagnostics.ts
Razzeee: Agreed, filter is correct, but why only add the first item to neededImports?

jmbockhorst: Because `neededImports` is for the "Import all" code action, and we only want to import the first suggestion.

Razzeee:
This is still acting weird for some cases that I would expect it to handle. 
Removing these two lines for e.g. https://github.com/rtfeldman/elm-spa-example/blob/master/src/Article/Feed.elm#L17-L18

Part of it seems to be due to fullyqualifed identifieres

src/providers/diagnostics/elmMakeDiagnostics.ts
```diff
@@ -76,14 +82,57 @@ export class ElmMakeDiagnostics {
...
```
Razzeee: text can be something like `Route.href`, which fails to find anything.

src/util/importUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+  }[] {
+    const exposedValues: {
+      module: string;
+      value: string;
+    }[] = [];
+
+    // Find all exposed values that could be imported
+    if (forest) {
+      forest.treeIndex.forEach((tree) => {
+        tree.exposing?.forEach((exposed) => {
```
Razzeee: If this is a `type`, there is a `exposedUnionConstructors` array, we need to handle too.

Razzeee:
Yeah, seems like I broke some of this with https://github.com/elm-tooling/elm-language-server/pull/227

src/providers/diagnostics/elmMakeDiagnostics.ts
Razzeee: Ahh, that makes sense, but now I'm wondering if we should sort these somehow at somepoint. Probably a task for later.

jmbockhorst: Yeah, I just thought about sorting and already implemented it using `RANKING_LIST`.

jmbockhorst:
I think fully qualified imports and union types are working good now.

Razzeee:
Looks great so far, still some odd cases here and there, will have to play with it a bit.

src/providers/diagnostics/elmMakeDiagnostics.ts
```diff
@@ -76,14 +82,73 @@ export class ElmMakeDiagnostics {
...
+            },
+          );
+
+          // Find imports
+          if (valueNode) {
+            exposedValues
+              .filter(
+                (exposed) =>
+                  exposed.value === valueNode.text ||
+                  (valueNode.type === "upper_case_qid" &&
```
Razzeee: There seem to be cases where `value_qid` is what we want here (code hereafter needs changes too)

Remove https://github.com/rtfeldman/elm-spa-example/blob/master/src/Article/Feed.elm#L17 and try to import `Profile.avatar` in line 132

Razzeee: But seems like #227 needs fixing before you can test this

Razzeee: Should work with master now.

jmbockhorst: Sweet, I updated it to handle the `value_qid` case as well.

Razzeee:
Hrm, lastest changes seem to have it broken completly. Can't get one suggestion to pop up.

Razzeee:
So another case we should handle is, adding imports, when there are no imports yet. Either going for the module declaration or if that is missing just the top of the file.

jmbockhorst:
Yep, thanks for catching that.

Razzeee:
Thank you, I'll merge this for now, we probably want to keep an eye out for edgecases. Awesome work so far. :unicorn:

PR #249, opened 2020-04-03
Record definition and completion fixes

### LINKED ISSUE #246: Jump to definition for record fields not working
"Jump to definition" doesn't seem to work for record fields in 1.7.0 in VSCode or in Vim.

Using the test file below, right-clicking and pressing "Go To Definition" on any of the instances of `fieldNumberOne` doesn't jump to the definition in the type alias at the top.

```elm
module Test exposing (..)


type alias TestType =
    { fieldNumberOne : Int
    , fieldNumberTwo : Int
    }


testValue : TestType
testValue =
    { fieldNumberOne = 1
    , fieldNumberTwo = 2
    }


testField : TestType
testField =
    let
        jumpTest1 =
            testValue.fieldNumberOne

        jumpTest2 =
            { testValue | fieldNumberOne = 3 }

        jumpTest3 : TestType
        jumpTest3 =
            { fieldNumberOne = 4, fieldNumberTwo = 3 }
    in
    testValue
```

VSCode version 1.43.2

jmbockhorst:
Fixes #246 as well as record completions for some imported functions. 

**Technical summary of changes:**
The function `getTypeAliasOfRecord` makes two definition jumps to find the type, and if the first jump is to another file, the next jump need to have the correct `uri` and `tree`. To do this, I had to pass through the `forest` so it can find the correct `tree` after the first jump. I also had to pass it into `findDefinitionNodeByReferencingNode`, but made it optional since its only needed for this specific case. Also, `getTypeAliasOfRecord` now returns the `uri` as well, so any further uses will know where the type alias is.

### DIFF
--- a/src/util/treeUtils.ts
+++ b/src/util/treeUtils.ts
@@ -1,6 +1,7 @@
 import { Position } from "vscode-languageserver";
 import { SyntaxNode, Tree } from "web-tree-sitter";
 import { IImport, IImports } from "../imports";
+import { IForest } from "../forest";
 
 export type NodeType =
   | "Function"
@@ -658,6 +659,7 @@ export class TreeUtils {
     uri: string,
     tree: Tree,
     imports: IImports,
+    forest?: IForest,
   ): { node: SyntaxNode; uri: string; nodeType: NodeType } | undefined {
     if (
       nodeAtPosition.parent &&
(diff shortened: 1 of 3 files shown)


Razzeee:
Thanks for looking into this

src/util/treeUtils.ts
```diff
@@ -1386,21 +1439,26 @@ export class TreeUtils {
             );
           } else if (definitionNode.nodeType === "TypeAlias") {
             aliasNode = definitionNode.node;
```
Razzeee, line 1441: hrm, the logic is a bit weird now. and we can drop this variable assignment, as it has become unused.

jmbockhorst: Yeah, I forgot to remove it in this commit but I have in my next set of changes that I’m working on.

Issue #247, opened 2020-04-02
Autocomplete for record access


bengolds:
This is a feature request.

Based off of the "autocomplete for record updates" feature in 1.7.0, it'd be nice to see the same feature for record access. e.g.

```elm
bob = 
  { name = "Bob"
  , age = 32
  , address = "Nowhere"
  }

shipTo =
  bob.|
     -- Would be nice to see autocomplete show name, age, and address here.
```

I'm not sure if this would require more type annotation than the record update feature -- but it'd be handy!

jmbockhorst:
@Razzeee is there something in the client grammar that needs updated for this to work? I can get the server to send the correct completions to the client, but VSCode still isn’t showing them.

Razzeee:
That shouldn't be the case, are you sending them as `bob.name` or as `bob`? The range field on the completion might also cause problems I guess?

jmbockhorst:
I was just sending `name` with the range starting after the `.`. Wouldn't that be the ideal completion?

Razzeee:
Might be, but you probably need to change the range then, from memory I would think that the range expects it to be `bob.name`. 

It's kind of a hack, as I build completions this way to be able to complete stuff not based on sections, but complete stuff. So you will write `Htm` and get `Html.Attributes.class` in the list, most other languages would just complete `Html`, let you enter a point and then figure out the next section. I mostly did it the other way, as it seemed easier.

bengolds:
This works great in the newest version! 🙌

PR #251, opened 2020-04-03
Add record access completions for types and nested types

### LINKED ISSUE #237: Completions for nested types
There should be completions for nested types. For example:

``` elm
type alias Data =
    { prop : String
    }


type alias Model =
    { data : Data
    }


test : Model -> String
test model = model.data.|
```

`model.data.` should have completions for the `Data` type.

jmbockhorst:
Closes #237. Completions for record access of types and nested types. 

Part of #247. For this case to work, `getReturnTypeOrTypeAliasOfFunctionDefinition` needs to be updated to handle anonymous type alias return types, but that example should work if the function had a type annotation.

### DIFF
--- a/src/providers/completionProvider.ts
+++ b/src/providers/completionProvider.ts
@@ -166,6 +166,29 @@ export class CompletionProvider {
           params.textDocument.uri,
           forest,
         );
+      } else if (
+        (nodeAtPosition.type === "field_access_segment" ||
+          nodeAtPosition.parent?.type === "field_access_segment") &&
+        nodeAtPosition.parent
+      ) {
+        const dotIndex = (
+          TreeUtils.findFirstNamedChildOfType("dot", nodeAtPosition) ??
+          TreeUtils.findFirstNamedChildOfType("dot", nodeAtPosition.parent)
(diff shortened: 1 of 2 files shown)


Razzeee:
This seems to only work in some cases.

If I take your example
```
test : Model -> String
test model = model.data.|
```
and just press Control+Space I'm not getting (good) recommendations

Razzeee:
Did some more tests and I think we need to do some work on the AST to make this work good.

For e.g.:
"field_access_expr" seems to wrongly match this string

"model.


type"

https://github.com/Razzeee/tree-sitter-elm

jmbockhorst:
Yeah, I noticed that and basically did some hacks to get it working properly for me, but it would be better to get the AST correct.

Razzeee:
Not sure how to move this along, I've been trying to get the tree to break like I want it to break, but there's no solution I was able to figure out that covered 100%

I might want to work on something else for a change.

https://github.com/tree-sitter/tree-sitter/issues/590

Razzeee:
I'll merge this for now, as it still is an improvement.

PR #255, opened 2020-04-17
Fix elm.json ignored when path similar to another

sentience:
When the path of one elm.json file matches the start of another, the
second one is ignored. This works as intended when the second path is
a nested directory:

- /path/to/parent/elm.json
- /path/to/parent/child/elm.json - this should be ignored

This change fixes a bug where we were unintentinally ignoring elm.json
files in directories whose names started with the same characters as a
sibling:

- /path/to/foo/elm.json
- /path/to/foobar/elm.json - this should not be ignored

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -210,12 +210,12 @@ export class Server implements ILanguageServer {
 
   private findTopLevelFolders(listOfElmJsonFolders: URI[]) {
     const result: Map<string, URI> = new Map();
-    listOfElmJsonFolders.forEach(uri => {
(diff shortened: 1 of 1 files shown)


Razzeee:
Nice find!
Can you please run prettier on this? I'm pretty sure it has something to complain about :)

sentience:
Good catch! Looks like I didn't **npm install** and so I wasn't running Prettier 2.0 (which changed the default for [arrow function parens](https://prettier.io/docs/en/options.html#arrow-function-parentheses)).

PR #260, opened 2020-04-24
add emacs doom configuration

ni-ko-o-kin:


README.md
```diff
@@ -224,6 +224,43 @@ elmAnalyseTrigger = "change"
 
 The language client is included in [lsp-mode](https://github.com/emacs-lsp/lsp-mode), specifically [here](https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-elm.el). See specifically [this section](https://github.com/emacs-lsp/lsp-mode#use-package) for a minimal use-package configuration for lsp-mode.
 
+#### Emacs Doom
+
+- Enable these layers in your configuration file `.doom.d/init.el` (`SPACE f P`):
```
Razzeee, line 229: As a total emacs noob, I'm not sure what you want me to. Append the lines below to the file?

README.md
```diff
@@ -224,6 +224,43 @@ elmAnalyseTrigger = "change"
...
+lsp
+(elm
+ +lsp)
+```
+
+- Config [lsp-ui](https://github.com/emacs-lsp/lsp-ui) features (peek, doc, sideline, etc) at `.doom.d/config.el`:
+
+```elisp
+(after! lsp
+  (setq lsp-enable-symbol-highlighting nil)
```
Razzeee, line 241: Can you explain what these are about? Can we maybe add that in comments?

ni-ko-o-kin: It's an example configuration, so people know how to configure lsp in case they are new to emacs.

I could add "Example configuration" above the code.

README.md
ni-ko-o-kin: This is an example for such a file https://github.com/hlissner/doom-emacs/blob/develop/init.example.el

"To enable" means, you have to uncomment elm and lsp, and add +elm to the lsp layer.

What about rephrasing it like this: Uncomment or add these lines in your configuration file ...

Razzeee: I think changing the example to be two lines (https://github.com/hlissner/doom-emacs/blob/develop/init.example.el#L123) would make sense and directly refer to them.
Uncomment or add the `lsp` and the `(elm +lsp)` line in your config.

README.md
Razzeee: So this part is completly optional but improves the experience?

ni-ko-o-kin: Yes! Especially the max-width and max-height. The popup is unusable with its default configuration.

README.md
ni-ko-o-kin: In the default init.el file, which is included when you install emacs doom (so no need to reference the init.example.el file), lsp and elm are both commented.

Razzeee: Oh, so you just need to uncomment `elm` and not have `(elm +lsp)` at all?

ni-ko-o-kin: no, you have to add +lsp as well... I will mention it.

Razzeee:
Thank you! Looks good and understandable to me :)

:tada:

PR #267, opened 2020-05-04
Remove exposing subscriptions in Browser.sandbox snippet

wingyplus:
Browser.sandbox didn't have subscriptions function. When use this
snippet, Elm will complain about function not found in exposing clause.


Razzeee:
Good catch, thank you!

PR #271, opened 2020-05-17
Sort auto imports by closest similar modules

### LINKED ISSUE #268: Auto import should prefer modules in the same folder
Auto imports should be sorted based on how close the module is to the current one. Modules in the same folder should be preferred over others.

jmbockhorst:
Fixes #268.


Razzeee:
Thanks for looking into this.

PR #272, opened 2020-05-17
Handle some more record field access cases

jmbockhorst:
Fix some more record access cases. Specifically it fixes this one, but it may improve more areas too:

``` elm
view : Model -> Html Msg
view model =
    let
        var : String
        var =
            model.|
                |> String.toFloat
                |> String.fromFloat
    in
    div [] []
```

What do you think about trying to write test cases for these? It would make it a lot easier to know what it working and to not break stuff in the future. I assume that will involve a lot to set up a mock workspace and forest, etc.

### DIFF
--- a/src/providers/completionProvider.ts
+++ b/src/providers/completionProvider.ts
@@ -168,17 +168,29 @@ export class CompletionProvider {
         );
       } else if (
         (nodeAtPosition.type === "field_access_segment" ||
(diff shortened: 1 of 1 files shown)


Razzeee:
Tests would be great. I even considered writing an additional project in haskell, as they seem to have quiet a nice lsp test framework https://hackage.haskell.org/package/lsp-test
But I never used haskell for anything.

PR #273, opened 2020-05-18
Add tests for the completion provider

jmbockhorst:
This is a WIP to add tests for the completion provider (and support for other providers). Let me know what you think of the direction of this so far.


src/providers/completionProvider.ts
```diff
@@ -20,25 +20,30 @@ import { HintHelper } from "../util/hintHelper";
...
 
-  constructor(private connection: IConnection, elmWorkspaces: ElmWorkspace[]) {
-    connection.onCompletion(
-      new ElmWorkspaceMatcher(elmWorkspaces, (param: CompletionParams) =>
-        URI.parse(param.textDocument.uri),
-      ).handlerForWorkspace(this.handleCompletionRequest),
-    );
+  constructor(
+    private connection?: IConnection,
+    elmWorkspaces?: ElmWorkspace[],
```
Razzeee, line 31: Should probably be `IElmWorkspace` too, for consistency's sake

src/providers/completionProvider.ts
```diff
@@ -20,25 +20,30 @@ import { HintHelper } from "../util/hintHelper";
...
   private qidAtStartOfLineRegex = /^[a-zA-Z0-9 \.]*$/;
 
-  constructor(private connection: IConnection, elmWorkspaces: ElmWorkspace[]) {
-    connection.onCompletion(
-      new ElmWorkspaceMatcher(elmWorkspaces, (param: CompletionParams) =>
-        URI.parse(param.textDocument.uri),
-      ).handlerForWorkspace(this.handleCompletionRequest),
-    );
+  constructor(
+    private connection?: IConnection,
```
Razzeee, line 30: This is creating an empty constructor? To help making this testable?

src/test/completionProvider.test.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+}
+
+describe("CompletionProvider", () => {
+  const completionProvider = new MockCompletionProvider();
+  const treeParser = new SourceTreeParser();
+
+  async function testCompletions(
+    source: string[],
+    line: number,
```
Razzeee, line 22: Might be nicer to use an object that has both line and character. I'm also not sure why you always add `+ 2` so that might need some docs.

Ultimately, I would try to move this syntax to something that's compatible with the one intellij-elm is using

```
--@ main.elm
g = User.defa{-caret-}
--@ User.elm
module User exposing (..)
defaultUser = "Arnold"
""", """
g = User.defaultUser{-caret-}
```

`{-caret-}` defines where your typing, I don't think lsp/vscode let you have multiple of these in the completion case.
`--@ name.elm` defines a file.

https://github.com/klazuka/intellij-elm/blob/master/src/test/kotlin/org/elm/lang/core/completion/ElmCompletionTest.kt

src/providers/completionProvider.ts
jmbockhorst: Yes, because testing doesn't need an `IConnection`. It's not a great solution.

Razzeee: Can't jest just create a mock for you? I'm not sure if that's how it works.

jmbockhorst: Not sure, will look into it.

src/providers/completionProvider.ts
```diff
@@ -20,25 +20,30 @@ import { HintHelper } from "../util/hintHelper";
...
   }
 
-  private handleCompletionRequest = (
+  protected handleCompletionRequest = (
     params: CompletionParams,
-    elmWorkspace: ElmWorkspace,
+    elmWorkspace: IElmWorkspace,
   ): CompletionResult => {
-    this.connection.console.info(`A completion was requested`);
+    this.connection?.console.info(`A completion was requested`);
```
jmbockhorst, line 46: @Razzeee ultimately we have to call `handleCompletionRequest` and this line will fail if we don't have the check if connection is undefined. So I'm not sure a mock would do anything with this in here.

Razzeee: I think a syp/mock from jest might be able to not fail here. I can try to have a look later this evening, if you don't beat me to it.

Razzeee: I pushed a commit, please have a look.

jmbockhorst: Looks great! Thanks for looking into it.

jmbockhorst:
If you think it’s ready, we could probably merge this now as a good base and then work on adding more tests.

Razzeee:
Thank you for working on this, I think we're in a good place to extend this now.

PR #277, opened 2020-05-23
Add some completion tests

jmbockhorst:
Just add some simple completion tests. The default test is just checking that the completions are somewhere in the list (Since the client does the sorting). Some of the tests check that only those completions are in the list (Like record access) using the `testExactCompletions` flag.


Razzeee:
While the client technically sorts, we are providing a sort order. So that could be tested, not saying you should do that, just pointing out, that your observation is not 100% true.

PR #280, opened 2020-05-27
Add more completion provider tests

jmbockhorst:
Fixed a major silly bug in completion tests (`.toBeTruthy` -> `.toBeTruthy()`) and fixed appropriate tests. Added more tests to cover most of the cases. Here is a summary of the code coverage improvements (statements). 

`completionProvider.ts` 60% -> 91%
`hintHelper.ts` 59% -> 77%
`treeUtils.ts` 41% -> 48%


src/test/utils/mockElmWorkspace.ts
```diff
@@ -1,21 +1,28 @@
...
 import { URI } from "vscode-uri";
-import Parser from "web-tree-sitter";
+import Parser, { Tree } from "web-tree-sitter";
 import { IElmWorkspace } from "../../elmWorkspace";
 import { Forest } from "../../forest";
 import { Imports } from "../../imports";
+import { readFileSync } from "fs";
 
 export const mockUri = Path.join(__dirname, "../sources/src/Test.elm");
+export const mockUri2 = Path.join(__dirname, "../sources/src/Test2.elm");
```
Razzeee, line 10: FYI, I think we need this configurable. And I would like to explore an api like https://github.com/klazuka/intellij-elm/blob/master/src/test/kotlin/org/elm/lang/core/resolve/ElmLayeredImportResolveTest.kt#L10-L24

Which doesn't mean that this is what's passed to the constructor, but it might be.

Razzeee:
Looks good otherwise. Implementing muliple files like the intellij-elm tests do, will also enable us to match their tests. Not sure I like their completion tests, as it only seems to be testing for one value? But it might at least be an inspiration https://github.com/klazuka/intellij-elm/tree/master/src/test/kotlin/org/elm/lang/core/completion and base for similar cases.

src/test/utils/mockElmWorkspace.ts
jmbockhorst: Can you explain what you mean more? I’m not sure I understand what api you are referring to in that file.

Razzeee: The string 

```
  """
--@ main.elm
import Foo as F
import FooExtra as F
main = F.bar
         --^Foo.elm
--@ Foo.elm
module Foo exposing (..)
bar = 42
--@ FooExtra.elm
module FooExtra exposing (..)
quux = 99
"""
```

will be parsed into three files, main.elm, Foo.elm and FooExtra.elm that's quiet a flexible approach

jmbockhorst:
@Razzeee Added the ability to parse multiple files using the `--@ File.elm` syntax and adopted that in the completion tests. All the completion tests should now be working.

src/test/completionProvider.test.ts
```diff
@@ -80,275 +82,377 @@ describe("CompletionProvider", () => {
...
-      `  { prop1: String`,
-      `  , prop2: Int`,
-      `  }`,
-      ``,
-      `view : Model -> Model`,
-      `view model =`,
-      `  { model | p{-caret-} }`,
-    ];
+    const source = `
+--@ Test.elm
```
Razzeee, line 86: Do we now depend on this being here? I would like this to assume it's file X, when we don't have one of the entries. That way we could omit these and basically copy these strings from the intelliJ tests

src/test/definitionProviderTests/definitionProviderTestBase.ts
```diff
@@ -2,10 +2,13 @@ import { DefinitionProvider, DefinitionResult } from "../../providers";
...
 import { IElmWorkspace } from "../../elmWorkspace";
 import { SourceTreeParser } from "../utils/sourceTreeParser";
-import { mockUri } from "../utils/mockElmWorkspace";
+import { baseUri } from "../utils/mockElmWorkspace";
 import { mockDeep } from "jest-mock-extended";
 import { TreeUtils } from "../../util/treeUtils";
 import { getInvokeAndTargetPositionFromSource } from "../utils/sourceParser";
+import { URI } from "vscode-uri";
+
+const mockUri = URI.file(baseUri + "Main.elm").toString();
```
Razzeee, line 11: I'll need to work on this next (this is a note for me)

src/test/completionProvider.test.ts
jmbockhorst: No, I should still parse it as `Main.elm` if it doesn't see any `--@ xxx.elm` definitions.

PR #281, opened 2020-05-28
Added more definition tests and more example parsing

Razzeee:
@jmbockhorst 

Please have a look, this is doing some of the parsing we need for multiple files in one inline string.


src/test/definitionProviderTests/definitionProviderTestBase.ts
```diff
@@ -0,0 +1,107 @@
...
+            this.treeParser.getWorkspace(determinedTestType.sources),
+          );
+
+          expect(definition).toEqual(undefined);
+        }
+        break;
+
+      case "resolvesToDifferentFile":
+        {
+          const definition = this.definitionProvider.handleDefinition(
```
Razzeee, line 55: This branch is not used yet and needs changing

src/test/utils/sourceParser.ts
```diff
@@ -2,63 +2,137 @@ import { Position } from "vscode-languageserver";
...
-      result.position = { line, character };
+      tempPosition = { line, character };
     }
   });
 
+  if (!tempPosition) {
+    fail();
+  }
+
+  tempSource = ["module Test exposing (..)", "", ...tempSource];
```
Razzeee, line 22: I think we should remove this and do it explicitly in every test string

src/test/utils/sourceParser.ts
```diff
@@ -2,63 +2,137 @@ import { Position } from "vscode-languageserver";
...
+
+  (tempPosition as Position).line += 2;
+
+  const result: {
+    newSources: { [K: string]: string };
+    position: Position;
+  } = { newSources: {}, position: tempPosition };
+  result.newSources["Main.elm"] = tempSource.join("\n");
```
Razzeee, line 30: This needs to change to handle multiple files. But it should be easy from here.

jmbockhorst:
I was actually in the middle of working on it too, but I wasn't too far into yet.

src/test/utils/sourceParser.ts
jmbockhorst: Yeah, I was thinking that too.

Razzeee:
Thank god, sorry for doing this unannounced. :)

Razzeee:
I would be willing to merge it as is, if you agree on the direction we're going. 

Will do some follow ups in the next days to ignore the failing tests and add the missing handling for jump to different file tests.

You could work on the completions side of things then.

jmbockhorst:
I'll wait until this is merged and pull it into the other branch and use it to parse multiple files.

PR #285, opened 2020-06-01
Enable a few completions tests that already work

jmbockhorst:
Just enabling a few tests based on the current behavior.

### DIFF
--- a/src/test/completionProvider.test.ts
+++ b/src/test/completionProvider.test.ts
@@ -573,7 +573,7 @@ func =
     ]);
   });
 
-  xit("Module name with caret after dot should have completions", async () => {
+  it("Module name with caret after dot should have completions", async () => {
     // TODO: Add auto import completions for modules and module values
     const source = `
 --@ Data/User.elm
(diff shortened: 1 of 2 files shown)


src/test/completionProvider.test.ts
```diff
@@ -592,7 +592,7 @@ test =
   Data.{-caret-}
```
jmbockhorst, line 592: In the future, do we want this be an exact completion list of just `User` in this case, since its the only thing that should be after the dot? That's how I would prefer, but I know that's not how you've currently set it up.

src/test/completionProvider.test.ts
```diff
@@ -613,10 +613,15 @@ test =
   Data.User.{-caret-}
```
jmbockhorst, line 613: This would be just `func` and `TestType`

src/test/completionProvider.test.ts
```diff
@@ -630,7 +635,7 @@ defaultPage =
   Page.{-caret-}
```
jmbockhorst, line 635: This would be just `Home` and `Away`.

src/test/completionProvider.test.ts
Razzeee: Sure, it just was easier to setup back then. I also thought, it might help with discovery especially for newer folks, as you will get the same function, but imported differently. If I remember right.

Changing this, might also help with the problems this decision caused due to some clients not being able to handle this correctly.

PR #290, opened 2020-06-12
Test FDir

Razzeee:


sentience:
Works without issue or noticeable performance degradation on our large project containing 30 different elm.json files. The log output during startup seems identical, the functionality all seems to work, and the indexing of files completes in about 30 seconds.

Razzeee:
This only adds additional logging, so that we can see if Globby and FDir act the same way.

sentience:
Oh! I'll take another look. :)

sentience:
OK, the two lists of files for each of the globs are the same, but appear in a different order. At a glance, it looks like one of them is doing a depth-first search, and the other is doing a breadth-first search – or something similar. Last file listed by Globby is often the first file listed by Fdir.

Knowing how the language server works internally, I'm guessing the order of the files is not important, and so a change from Globby to Fdir is likely to be fine.

Razzeee:
That's great. Order shouldn't matter, as you said.

PR #289, opened 2020-06-11
Fix imports being skipped due to early return

Razzeee:
Closes #288 

@jmbockhorst Probably not the best solution, but hopefully good enough for a hotfix.

Issue #288, opened 2020-06-11
Auto complete for external packages (including core) have stopped working for me since the release of 1.0.0


Kurren123:
[Simple sample repository](https://github.com/Kurren123/elm-vscode-bug). 

I clone this, run `elm make` and it compiles just fine. Autocomplete works fine for functions of modules defined in this project, but not for functions in other modules such as `Html`:

![image](https://user-images.githubusercontent.com/3525980/84401586-5a660d80-abfb-11ea-94de-25dea777b703.png)

## My Environment
* Version used: 1.0.0
* Editor name and version: VSCode 1.46
* Environment name and version:

![image](https://user-images.githubusercontent.com/3525980/84402175-07d92100-abfc-11ea-94a1-6dc74c55b9a9.png)

* Operating System and version: Windows 10 Enterprise 64 bit

Please let me know if there is any more info you need

Kurren123:
@Razzeee  Thanks for the update but the error seems like it's still there even after updating the extension to v1.0.1 and restarting vscode. 

![image](https://user-images.githubusercontent.com/3525980/84494674-d581ff00-aca1-11ea-915f-4f65c450f0cb.png)

There are more autocomplete suggestions than before so something did change, but still none from `Html`.

Razzeee:
Seems fine to me on linux, will test on windows shortly

![image](https://user-images.githubusercontent.com/5943908/84495134-f3a02d00-acaa-11ea-84c4-cb0303a83759.png)

Razzeee:
It's a problem on windows due to file paths being handled wrong. Sorry, will push another update :)

Kurren123:
No problem at all, I'm just grateful for the bugfix

Issue #287, opened 2020-06-09
Reduce weight of auto-import suggestions


sparksp:
## Expected Behaviour

If an already imported module provides a function (without `exposing` it) then the qualified name should be suggested before the auto-import suggestions.

If the value being suggested has already been used elsewhere in the file I expect the top suggestion to match that use. For example, if I have already used `Set.member` in the file then typing `member` should present `Set.member` above the auto-import suggestion (assuming that `member` has not been exposed on import).

## Current Behaviour

Continuing with the previous example, if there is no `member` in scope (declared or exposed on import) then the top suggestion(s) for "member" are to auto-import it from various modules; the available fully-qualified suggestions then follow. In the case of `member` there's up to 6 possible auto-imports to skip over.

If `member` has been exposed on import already then calling it is the top suggestion, followed by auto-import suggestions for the remaining modules (up to 5), and finally the fully-qualified suggestions.  This seems to suggest an additional problem, since you wouldn't want to auto-import `member` from more than one module in the same file.

## Possible Solution

Suggestions should favour things that are already available over auto-import, for example:

1. In-scope named value (let value, case value, function parameter, top-level declaration, etc.)
2. Exposed value from imported module (e.g., `member` from `import Set exposing (member)`)
3. Qualified values from imported module (e.g., `Dict.member` from `import Dict` and `List.member` from default `import List`)^
4. Auto-import suggestions

^ Further weighting could be applied to (3) based on scope and whether the qualified value appears in the file already; for example showing `List.member` above `Dict.member` if `List.member` has already been used in the file; or showing the best match if we're following `something |>`.

## Context

I've been writing [elm-review] tests where I `import Review.Test` and then fully qualify function calls, including `Review.Test.expectErrors`, `Review.Test.atExactly` and `Review.Test.whenFixed`. Previously it was sufficient to start typing `expectErrors` or `atExactly` and accept the top suggestion to get the fully qualified name.

When importing I typically only expose types from modules that match the module name, and fully-qualify everything else, so I'm unlikely to want auto-import to expose functions for me.  For example I `import Html exposing (Html)` and then use `Html.div` later; or `import Review.Rule as Rule exposing (Rule)` and use `Rule.*` later.

[elm-review]: https://package.elm-lang.org/packages/jfmengels/elm-review/latest/

## Your Environment

* Version used: elmtooling.elm-ls-vscode 1.0.0
* Editor used: VSCode 1.45.1
[...]

PR #291, opened 2020-06-17
Completion improvements

jmbockhorst:
### Improved sorting of qualified imports and auto imports. Ref #287.
The problem is that if you have `Html` imported and you type `div`, there are 2 completions: `div (auto imports)` and `Html.div`. The auto imported will always be first because it matches the string better (in VSCode at least). To fix this, I check if the typed text matches the function name and add filter text to `Html.div` completion. So if we type `di`, `Html.div` gets a filter text of `div`, and thus will sort to the top. If we start to type `Ht`, the filter text will just be the label, `Html.div`. This is the main fix for #287, but there are still some cases I want to look at later.

### Added completions for module values or possible submodules
If a module is imported, it now should all possible values for that module. Ex: `Html.{-caret-}` will show `a, div, span, ...`. It also shows possible submodules (not imported). In the case: `Attributes` or `Styled.Attributes` could also show up as completions. 
 
### Fixed record completion interfering with module completions. Ref #288.   
Reverts #289 with a proper fix. Instead or checking for all `field_access_segment` descendents, it now specifically looks at the target word. 

I also refactored `CompletionProvider` to use object parameters instead of function parameters. It made it more clear and easier to pass in optional parameters (like `filterText`).

And of course, added a lot of tests to cover these cases.

### DIFF
--- a/src/providers/completionProvider.ts
+++ b/src/providers/completionProvider.ts
@@ -188,30 +188,42 @@ export class CompletionProvider {
           forest,
         );
       } else if (
-        (nodeAtPosition.type === "field_access_segment" ||
-          nodeAtPosition.parent?.type === "field_access_segment" ||
-          TreeUtils.descendantsOfType(nodeAtPosition, "field_access_segment")
-            .length > 0) &&
-        nodeAtPosition.parent
+        nodeAtPosition.type === "field_access_segment" ||
+        nodeAtPosition.parent?.type === "field_access_segment" ||
+        RegExp("^[a-z]+.*.$").exec(targetWord)
       ) {
-        const accessSegmentNode =
+        let accessSegmentNode =
           nodeAtPosition.type === "field_access_segment"
             ? nodeAtPosition
             : nodeAtPosition.parent?.type === "field_access_segment"
             ? nodeAtPosition.parent
-            : TreeUtils.descendantsOfType(
-                nodeAtPosition,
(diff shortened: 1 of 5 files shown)


jmbockhorst:
Also pushed a fix as a follow up to 71cad988fe31243606d8dbe4231023727f1836f6 where `element.replace(".", "/")` didn't replace all the dots. This caused paths like `Module/Submodule.AnotherSub`

Razzeee:
Do you know why `replace` isn't enough? I'm not sure I understand what's going on.

jmbockhorst:
Because it only replaces the first match. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace

Razzeee:
Oh crap thank you. Not sure who designed that api, but I would have never expected that.

src/providers/completionProvider.ts
```diff
@@ -219,11 +288,17 @@ export class CompletionProvider {
         ...this.findDefinitionsForScope(nodeAtPosition, tree, replaceRange),
       );
 
+      const currentTarget = targetLine.substring(
```
Razzeee, line 291: Isn't this identical to `targetWord`?

jmbockhorst: Yeah I think so. I added this first then targetWord later, so should be good to get rid of this one.

Issue #293, opened 2020-06-17
Request textDocument/completion failed with message: Invalid regular expression


predragnikolic:
While update the server from version 1.6.3 to 1.8.0 
I noticed the the completion request is not working correctly.

I can reproduce the problem with elm-language-server 1.8.0 and 1.8.3

Steps to repro. 
1. clone https://github.com/rtfeldman/elm-spa-example
2. open `/elm-spa-example/src/Api.elm`
3. When triggering completions on certain places in the document(see the gif). This elm server will throw an error:
```
[...]

PR #294, opened 2020-06-18
Don't complete in comments

Razzeee:


Razzeee:
@jmbockhorst 
The secound commit tries to improve handling when https://github.com/elm-tooling/elm-language-server/issues/293 happens

Razzeee:
Seems to cause a different problem.

Razzeee:
@jmbockhorst this is the valid fix. let's escape our regex inputs from now on, if they use variables. :)

Issue #183, opened 2019-10-19
Missing autocomplete for used but not defined functions


lydell:
## Steps to Reproduce and Context

Starting code:

```elm
module Main exposing (..)

import Json.Encode


type alias Scope =
    { user : User
    }


type alias User =
    { name : String
    }


encodeScope : Scope -> Json.Encode.Value
encodeScope scope =
    Json.Encode.object
        [ ( "tag", Json.Encode.string "Scope" )
        , ( "user", encodeUser scope.user )
        ]
```

Imagine that you’re currently working on the `encodeScope` function. You have just typed `encodeUser scope.user`. Now it’s time to define `encodeUser`. (I often work this way, starting with something and defining helper functions as I go.)

After `encodeScope`, start making the new function. The intention is to write `encodeUser : User -> Json.Encode.Value`. After typing `enc` I’d want `encodeUser` to be one of the suggestions, but I only get `encodeScope`.

After typing the annotation, `encodeUser : User -> Json.Encode.Value`, it’s time to type the definition. Again, when typing `enc`, with the intention of typing `encodeUser user =` I only get `encodeScope` as a suggestion.

Once the function actually exists it finally appears in autocomplete.

A workaround is to copy `encodeUser` just after having typed `encodeUser scope.user` so that I can paste it when starting to create the function. But sometimes I forget. Then it’s nice to use autocomplete. It feels like I get autocompletion for undefined stuff in JS and TS, making it feel like I get _less_ autocomplete in Elm.

## Your Environment
* Version used: 0.6.0
* Editor name and version:  VSCode 1.39.2
* Environment name and version: Node.js v12.8.0
* Operating System and version: macOS Mojave

PR #295, opened 2020-06-21
Add function completion for used but not declared function

Razzeee:
https://github.com/elm-tooling/elm-language-server/issues/183

This needs tests


Razzeee:
@jmbockhorst FYI

PR #298, opened 2020-07-03
Add value completions for non-imported modules

jmbockhorst:
Trying to access a non-imported module like `Module.Sub.{-caret-}` will now show those possible values as completions and will auto import `Module.Sub` on completion. Previously this only worked on already imported modules.


Razzeee:
Thank you!

PR #299, opened 2020-07-12
Better completions context

jmbockhorst:
This PR begins to address the that problem that `nodeAtPosition` is often determined to be very different nodes based on similar positions. For example `var.|` and `var.pr|` can product vastly different results (lower level node vs higher level) and we have a lot of specific pattern matching to try to catch all of these cases (and the result is ugly). There are also many rarer variations that can occur (like in deeper nested statements) that we don't cover, and completions don't work there.

To solve this, I took heavy inspiration from the Typescript completion provider and introduce a new `contextNode` and a `TreeUtils.findPreviousNode()` to use instead of the existing `nodeAtPosition`. `contextNode` is the node right before the caret (but is also adjusted when typing after a dot). 

We can continue to use both nodes and transition some completions to use the new `contextNode` when and where it makes sense. Currently I am mostly just using it for completions that come after a dot. Doing this means I am able to consistently get the node we want completions on (before the dot, Ex: `Module.v|` -> `Module`, `model.|` -> `model`) no matter the surrounding code (I was impressed with the testing results of this). This also means that we can get rid of a lot of the ugly node type matching code, and use a more consistent, unified approach. This should also help get better results with some of those edge cases.

### DIFF
--- a/src/providers/completionProvider.ts
+++ b/src/providers/completionProvider.ts
@@ -23,6 +23,7 @@ import RANKING_LIST from "./ranking";
 import { ImportUtils } from "../util/importUtils";
 import { RefactorEditUtils } from "../util/refactorEditUtils";
 import { Utils } from "../util/utils";
+import { comparePosition, PositionUtil } from "../positionUtil";
 
 export type CompletionResult =
   | CompletionItem[]
@@ -88,7 +89,7 @@ export class CompletionProvider {
         currentCharacter--;
       }
 
-      const replaceRange = Range.create(
+      let replaceRange = Range.create(
         Position.create(params.position.line, currentCharacter),
         params.position,
       );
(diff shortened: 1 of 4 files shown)


Razzeee:
This probably superseeds what I tried to do with https://github.com/elm-tooling/elm-language-server/commit/a05078889a9036fb49e90b07292d6dd68f8a8a33#diff-21245dcc42cbe9ffe996a08b05568766R19

src/providers/completionProvider.ts
```diff
@@ -391,19 +386,35 @@ export class CompletionProvider {
   }
 
   private getImportableModules(
+    tree: Tree,
     forest: IForest,
     range: Range,
+    targetModule?: string,
   ): CompletionItem[] {
+    const currentModuleNameNode = TreeUtils.getModuleNameNode(tree);
+    console.log(currentModuleNameNode?.text);
```
Razzeee, line 395: Remove this for merging

Razzeee:
Looks good otherwise

jmbockhorst:
Thanks, should be good now.

Razzeee:
Thank you, I squashed this.

PR #302, opened 2020-07-31
Add definition handling for type variables

jmbockhorst:
This is needed for type inference that I'm working on, but I thought it would be best to minimize unrelated changes in that branch and merge this in now.


jmbockhorst:
Oh yeah, `flatMap` doesn't work on Node 10...

Razzeee:
Thank you, I pushed a small improvement and will merge it now.

PR #304, opened 2020-08-07
Resolve tree index on demand

Razzeee:


jmbockhorst:
We should think about how we want to resolve trees once we have type inference. We have 2 difference document `onChange` handlers, in the diagnostics provider and in the astProvider. For type inference, I added a type diagnostic provider to get diagnostics on document change. The problem was that I couldn't use the tree in the forest because the `astProvider` hadn't been called yet to update the tree. To fix this, I let the diagnostic provider own the `astProvider` so it would call it to update the tree when we got a document `onChange` in the diagnostics provider. Just wanted to give you a heads up that we will need to do something about this.

Razzeee:
Might be due to one listening to `TextDocumentEvents` and the other to `DocumentEvents`? Not very happy with that design anyway.

jmbockhorst:
Yeah I think that's the case.

Razzeee:
Superseeded by https://github.com/elm-tooling/elm-language-server/pull/354

PR #307, opened 2020-08-10
Improve definition for conflicting module/type names and various completion improvements

jmbockhorst:
Fixes definitions for conflicting names of types and modules. For example, a module import named `Model` and a type named `Model` conflicted before and broke things. To fix this, I detect if there is a dot after the name, then it is a module, or else its a type. 

There are also some various completions filtering and sorting improvements. Like if a module import is exposing values, we guess that the completions from that module should just be the value, without the module prefix. Also sort auto imports based on modules that are already imported.

### DIFF
--- a/src/util/treeUtils.ts
+++ b/src/util/treeUtils.ts
@@ -856,19 +856,33 @@ export class TreeUtils {
 
       let definitionFromOtherFile;
       if (!definitionNode) {
-        definitionFromOtherFile = this.findImportFromImportList(
-          uri,
-          TreeUtils.findImportNameNode(tree, upperCaseQid.text)?.text ??
-            upperCaseQid.text,
-          "Module",
-          imports,
-        );
-        if (definitionFromOtherFile) {
-          return {
-            node: definitionFromOtherFile.node,
-            nodeType: "Module",
-            uri: definitionFromOtherFile.fromUri,
-          };
+        // Make sure the next node is a dot, or else it isn't a Module
+        if (TreeUtils.nextNode(nodeAtPosition)?.type === "dot") {
+          const endPos =
+            upperCaseQid.text.indexOf(nodeAtPosition.text) +
+            nodeAtPosition.text.length;
+
+          const moduleNameOrAlias = nodeAtPosition.parent.text.substring(
+            0,
+            endPos,
+          );
+          const moduleName =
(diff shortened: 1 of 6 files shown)


src/util/treeUtils.ts
```diff
@@ -856,19 +856,33 @@ export class TreeUtils {
...
-            nodeType: "Module",
-            uri: definitionFromOtherFile.fromUri,
-          };
+        // Make sure the next node is a dot, or else it isn't a Module
+        if (TreeUtils.nextNode(nodeAtPosition)?.type === "dot") {
+          const endPos =
+            upperCaseQid.text.indexOf(nodeAtPosition.text) +
+            nodeAtPosition.text.length;
+
+          const moduleNameOrAlias = nodeAtPosition.parent.text.substring(
```
Razzeee, line 865: Are you sure there are really cases, where you can't just take the parents complete text?

src/util/treeUtils.ts
```diff
@@ -2050,4 +2064,22 @@ export class TreeUtils {
       node.parent?.parent?.firstNamedChild?.type === "import"
     );
   }
+
+  public static nextNode(node: SyntaxNode): SyntaxNode | undefined {
+    // Move up until we have a sibling
+    while (!node.nextNamedSibling && node.parent) {
```
Razzeee, line 2070: I haven't looked into the `walk` part of tree-sitter, but it's supposed to be pretty fast. Might be an option for code like this in the future.

Razzeee:
Seems good to be merged. But kinda curious why this isn't including some tests :)

src/util/treeUtils.ts
jmbockhorst: Well this allows for a qid like `Html.Attributes.xxxx`, if you go to definition with the cursor over `Attributes`, it'll go to the `HTML.Attributes` module instead of the `xxxx` value.

jmbockhorst:
Good point. I'll add a few tests.

src/util/treeUtils.ts
Razzeee: Ah nice, but one reason more for a small test :)

jmbockhorst:
Added some tests that all failed previously. Should be good to merge.

PR #310, opened 2020-08-22
Type inference

jmbockhorst:
Core type inference support for most language features besides Records. No need to merge this anytime soon, especially not before next release, but I wanted to get your feedback on what I have so far. 

We need to figure out what to do for diagnostics, because I don't think we can do type inference diagnostics on change and elm analyse on change at the same time. I think it would be good to add a setting to opt in to using type inference for on change diagnostics and other language features. Once it becomes stable enough, we can just turn it on by default. 

I went with the approach to map every `SyntaxNode` to an `Expression`, which gives better type control over the expression types and getting child nodes. This is really nice to work with, but my only concern is speed, which I'm not sure will be an issue or not. We certainly could do everything with just `SyntaxNode`s, but it would be more complex and less reliable. 

If you want to test it, you can uncomment the blocks in `hoverProvider` and `diagnosticsProvider`. Then you can hover over function declarations to see the inferred type and you will have diagnostics on change. The diagnostics are basic right now, but I think we should try to make them similar to the compiler. 

Feel free to contribute on what is there so far, especially around diagnostics and ast updating (Since they overlap with existing code). I will be working on Record support, so it shouldn't interfere.


src/util/types/typeInference.ts
```diff
@@ -0,0 +1,1829 @@
...
+    return this.ancestors.map((a) => a.bindings.get(e.text)).find(notUndefined);
+  }
+
+  public static valueDeclarationInference(
+    declaration: EValueDeclaration,
+    uri: string,
+    elmWorkspace: IElmWorkspace,
+    activeScopes: Set<EValueDeclaration>,
+  ): InferenceResult {
+    // TODO: Need a good way to get all visible values
```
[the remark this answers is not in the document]
jmbockhorst: In this case, `shadowableNames` are a list of names that will cause an error if they are defined again.

src/util/types/typeInference.ts
```diff
@@ -0,0 +1,1829 @@
...
+  message: string;
+}
+
+export interface InferenceResult {
+  expressionTypes: SyntaxNodeMap<Expression, Type>;
+  diagnostics: Diagnostic[];
+  type: Type;
+}
+
+export class InferenceScope {
```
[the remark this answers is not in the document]
jmbockhorst: I wouldn't, as its main class with the core type inference stuff. Maybe other stuff should move out of this file.

src/util/types/typeInference.ts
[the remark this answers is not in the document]
Razzeee: Good, did you take that naming from the reference implementation? Doesn't seem to be a very clear variable name?

src/util/types/typeInference.ts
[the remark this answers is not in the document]
Razzeee: Fine to me, I'm just used to linters screaming at you for doing this (C#) and reviewing this file was a bit painfull, due to the length :)

src/providers/foldingProvider.ts
```diff
@@ -75,6 +70,44 @@ export class FoldingRangeProvider {
...
+            });
+          }
+
+          if (letBody) {
+            folds.push({
+              endCharacter: node.endPosition.column,
+              endLine: node.endPosition.row,
+              kind: FoldingRangeKind.Region,
+              startCharacter: letBody.startPosition.column,
+              startLine: letBody.startPosition.row - 1,
```
Razzeee, line 98: I think the -1 is wrong?

src/providers/foldingProvider.ts
```diff
@@ -75,6 +70,44 @@ export class FoldingRangeProvider {
...
+            });
+          }
+        } else if (node.type === "if_else_expr") {
+          node.namedChildren.slice(1).forEach((child) => {
+            folds.push({
+              endCharacter: child.endPosition.column,
+              endLine: child.endPosition.row,
+              kind: FoldingRangeKind.Region,
+              startCharacter: child.startPosition.column,
+              startLine: child.startPosition.row - 1,
```
Razzeee, line 108: I think the -1 is wrong?

src/providers/foldingProvider.ts
jmbockhorst: I don't know what the behavior before was, but here was my thought process:
Before we folded on `in` node, which looked like this:
````
in
    value
    ...
````

Now it folds on the body part
```
value
...
```
So I subtract 1 from the start line to include the line with `in`, like before.

This is also what I did for if else. But again, I didn't check the current behavior of folding.

**Comment by Razzeee** [comment]
Fixed 6 of the tests 4 to go

**Comment by Razzeee** [comment]
While tests are green, I still suspect there to be problems, references seem to be borked for e.g.

File: src/util/treeUtils.ts [hunk 5]
```diff
@@ -481,15 +475,18 @@ export class TreeUtils {
...
-      ret = functions.find((elmFunction) => {
-        const declaration = TreeUtils.findFirstNamedChildOfType(
-          "function_declaration_left",
-          elmFunction,
-        );
-        if (declaration && declaration.firstNamedChild) {
-          return functionName === declaration.firstNamedChild.text;
-        }
-      });
+      ret = functions
```
Razzeee, line 478: This change seems to break references for functions

Razzeee:
Please check my latest code changes, as I'm not sure if they are unproblematic for type inference.

src/util/treeUtils.ts
jmbockhorst: This change was for definitions of functions to resolve to the function declaration left instead of the value declaration. This was how IntelliJ elm does it so it worked with the type inference setup. We could revert if you want though and fix the type inference, as that shouldn't be too hard.

Razzeee: I guess we could change how the references work too, will have to check if that breaks more. If that doesn't pan out, we could also just duplicate the method.

Razzeee:
References now seem fine, at least from some quick tests. I also took care to not cause problems with other implementations. We need to recheck the earlier changes.

src/providers/diagnostics/typeInferenceDiagnostics.ts
```diff
@@ -1,52 +1,157 @@
...
+    };
+  }
+
+  private getNodeRange(node: SyntaxNode): Range {
+    const end = PositionUtil.FROM_TS_POSITION(node.endPosition).toVSPosition();
+    return {
+      start: PositionUtil.FROM_TS_POSITION(node.startPosition).toVSPosition(),
+      end: {
+        ...end,
+        character: end.character + 1,
```
jmbockhorst, line 153: Is this supposed to be +1?

Razzeee: I think it was +1 before, not sure why. I think you added it when you created the typeInferenceDiagnostic file. Not sure if you based it on a copy?

Razzeee: Removed

src/providers/diagnostics/typeInferenceDiagnostics.ts
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  public createDiagnostics = (
+    tree: Tree,
+    uri: string,
+    elmWorkspace: IElmWorkspace,
+  ): Diagnostic[] => {
+    let diagnostics: Diagnostic[] = [];
+
+    const allTopLevelFunctions = TreeUtils.findAllTopLevelFunctionDeclarationsWithoutTypeAnnotation(
```
Razzeee, line 37: Should we do functions from let too?

src/util/treeUtils.ts
```diff
@@ -16,7 +16,8 @@ export type NodeType =
   | "AnonymousFunctionParameter"
   | "UnionConstructor"
   | "FieldType"
-  | "TypeVariable";
+  | "TypeVariable"
+  | "Port";
```
Razzeee, line 20: Can you add a test for the port case? https://github.com/elm-tooling/elm-language-server/issues/297

src/util/treeUtils.ts
```diff
@@ -661,15 +691,19 @@ export class TreeUtils {
...
+    ) {
       return { node: definitionNode, nodeType: "Type" };
     }
     definitionNode = this.findTypeAliasDeclaration(tree, nodeAtPosition.text);
     if (definitionNode) {
       return { node: definitionNode, nodeType: "TypeAlias" };
     }
     definitionNode = this.findUnionConstructor(tree, nodeAtPosition.text);
-    if (definitionNode) {
+    if (definitionNode && nodeAtPosition.parent?.type !== "") {
```
Razzeee, line 706: Do you remember what this is about? I don't think we should ever have an empty type?

src/providers/diagnostics/typeInferenceDiagnostics.ts
jmbockhorst: I would add a code action, but not a diagnostic. Since elm-analyse doesn't show a diagnostics it would cause a lot of additional diagnostics, and it seems to be pretty common to not do annotations inside lets. But a code action should still be available if wanted.

src/util/treeUtils.ts
jmbockhorst: Nope. Whoops.

src/providers/diagnostics/typeInferenceDiagnostics.ts
Razzeee: I partly agree. I agree, that we should only do a code action, but we shouldn't do that because elm-analyse does something/doesn't do something. In my vision elm-analyse is on it's way out, in fact I want to start writing our own implementation of most of it's functionality.

jmbockhorst: Yeah, I shouldn't have said the part about elm-analyse, as I could've made the same argument without it. (Since it's common to not have/want let annotations, we shouldn't show a diagnostic)

src/forest.ts
```diff
@@ -40,9 +40,11 @@ export class Forest implements IForest {
...
 
   public getExposingByModuleName(moduleName: string): IExposing[] | undefined {
-    const result = this.treeIndex.find(
-      (tree) => tree.moduleName === moduleName && tree.isExposed,
-    );
+    const result = this.treeIndex
+      .sort((x, y) => {
+        return x.isExposed === y.isExposed ? 0 : x.isExposed ? -1 : 1;
+      })
+      .find((tree) => tree.moduleName === moduleName);
```
jmbockhorst, line 47: Should be more performant to `filter((tree) => tree.moduleName === moduleName)` first, then sort, then take the first result.

PR #312, opened 2020-08-25
Improve module rename

Razzeee:


Razzeee:
Please review @jmbockhorst 

This should fix hopefully all module renamings. I only tested renaming from the source (need to head to bed).

Changing the folder via a rename is still broken/todo. But I think this is a significant enough improvement.

src/providers/renameProvider.ts
```diff
@@ -98,6 +147,23 @@ export class RenameProvider {
...
+    const sourceDir = elmWorkspace.getPath(file);
+
+    // The file is not in a source dir (shouldn't happen)
+    if (!sourceDir) {
+      return;
+    }
+
+    const newUri = `${sourceDir}/${moduleName.replace(".", "/")}.elm`;
+
+    return URI.parse(newUri);
```
jmbockhorst, line 164: I had to switch this to `URI.file` for it to work for me, which I think should be the proper thing to do.

src/providers/renameProvider.ts
```diff
@@ -98,6 +147,23 @@ export class RenameProvider {
...
+    file: URI,
+  ): URI | undefined {
+    const sourceDir = elmWorkspace.getPath(file);
+
+    // The file is not in a source dir (shouldn't happen)
+    if (!sourceDir) {
+      return;
+    }
+
+    const newUri = `${sourceDir}/${moduleName.replace(".", "/")}.elm`;
```
jmbockhorst, line 162: `String.prototype.replaceAll` is coming in ES2021, but for now we need to use `.split(".").join("/")` to replace all.

src/providers/renameProvider.ts
```diff
@@ -42,12 +45,44 @@ export class RenameProvider {
...
+    if (
+      affectedNodes?.originalNode.parent?.parent?.type === "module_declaration"
+    ) {
+      const fullModuleName = affectedNodes?.originalNode.parent.text;
+      const modulePrefix = fullModuleName.substring(
+        0,
+        fullModuleName.lastIndexOf("."),
+      );
+
+      newName =
```
jmbockhorst, line 66: Should we handle the case of renaming the middle or beginning part or a module? For example, in `Module.Sub.Main`, if I rename `Sub` or `Module`, it causes unexpected behavior by adding the new name to the end (replacing `Main` instead). I think we would have to handle the references rename for this too.

src/providers/renameProvider.ts
```diff
@@ -98,6 +147,23 @@ export class RenameProvider {
...
+    if (!sourceDir) {
+      return;
+    }
+
+    const newUri = `${sourceDir}/${moduleName.replace(".", "/")}.elm`;
+
+    return URI.parse(newUri);
+  }
+
   private getRenameAffectedNodes(
```
jmbockhorst, line 167: Not necessarily in this PR, but this should be pulled into a util so that we can use it in the `FileEventsHandler` to rename references when a file in renamed.

src/providers/renameProvider.ts
jmbockhorst: > Changing the folder via a rename is still broken/todo. But I think this is a significant enough improvement.

I just realized that this is probably what you were describing, sorry to bring it up again.

src/providers/renameProvider.ts
Razzeee: Will try, I remember having that first in one place, but needing to change it, due to something not matching otherwise. Will check again this evening.

src/providers/renameProvider.ts
Razzeee: You can also use the regex version of replace: `replace(/./, "/")`

Razzeee: But good catch

src/providers/renameProvider.ts
Razzeee: Correct, but your right, we should at least disable that case. In fact, I'm thinking if this might be some fallout of the '.' changes you did recently? (but don't worry about this)

But disabling is the easy way out for now, obviously, best would be to be able to handle this. I guess selecting the whole token for this rename might fix most of the problem, will try this evening.

src/providers/renameProvider.ts
jmbockhorst: I will test it again tonight, but I just remember that when I used `parse` for a uri like `C:\Users\...\Repos\elm-project\src/Module/Submodule/Main.elm`, it created a `URI` with the schema of `c`, and a uri of `/Users/...`, which didn't match anything.

Razzeee: Seems to still work fine for me with the change. The place where I did need to use `parse` was in line 74.

src/providers/renameProvider.ts
Razzeee: Turned out that hunch was right, pretty much simplifies everything.

src/providers/renameProvider.ts
Razzeee: Actually needs to be `replace(/\./g, "/")`

src/providers/renameProvider.ts
Razzeee: Moved it to a util, will you look into calling it from `FileEventsHandler`?

Razzeee:
Very happy with this now, my testing was brief, but only thing that I think could be improved are some cases where the affected nodes are not including all cases that are needed for a module rename (there seem to be some wrong places in there too). But shouldn't be handled in this PR.

src/providers/renameProvider.ts
jmbockhorst: Yep 👍

jmbockhorst:
Working great for me.

Issue #118, opened 2019-07-22
Make path to project configurable


Razzeee:
We need to know where your main files are, so that we can specifically start building these. 

At the moment, we always compile the currently opend file. This leads to errors not showing up and unreliable behavior.

We should still fallback to the current behavior if it is not set.

PR #315, opened 2020-08-30
Make entrypoints configurable via elm-tooling.json

Razzeee:
Closes https://github.com/elm-tooling/elm-language-client-vscode/issues/114 and https://github.com/elm-tooling/elm-language-server/issues/118

Thanks to @lydell for the definition of `elm-tooling.json` https://github.com/lydell/elm-tooling.json

### DIFF
--- a/src/providers/diagnostics/elmMakeDiagnostics.ts
+++ b/src/providers/diagnostics/elmMakeDiagnostics.ts
@@ -67,6 +67,41 @@ export interface IStyledString {
   string: string;
 }
 
+type NonEmptyArray<T> = [T, ...T[]];
+
+function elmToolingEntrypointsDecoder(json: unknown): NonEmptyArray<string> {
+  if (typeof json === "object" && json !== null && !Array.isArray(json)) {
+    if ("entrypoints" in json) {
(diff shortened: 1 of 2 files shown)


Razzeee:
We need to mention this in the docs, so that it's discoverable.

lydell:
I added more validation and logging: https://github.com/elm-tooling/elm-language-server/pull/316

src/providers/diagnostics/elmMakeDiagnostics.ts
```diff
@@ -666,7 +686,7 @@ export class ElmMakeDiagnostics {
                 // elm-test might supply absolute paths to files
                 file: errorObject.path
                   ? path.relative(cwd, errorObject.path)
-                  : relativePathToFile,
+                  : relativePathToFile[0],
```
lydell, line 689: Is it OK to only use the first one here? (I have no idea what the code is doing here, sorry)

Razzeee: I think this should never really happen and as long as the fallback in, we should be safe

PR #316, opened 2020-08-30
Validate elm-tooling.json more and add more logging

lydell:


README.md
```diff
@@ -85,6 +86,24 @@ npm install -g elm elm-test elm-format
 
 Or use local versions from your `node_modules` directory, if you want to do that you need to set the paths, via the settings (e.g. set `elmPath` to `./node_modules/.bin/elm`).
 
+## Configuration
+
+Create an [elm-tooling.json](https://github.com/lydell/elm-tooling.json) file next to your `elm.json` to configure the language server.
+
+Currently there’s just one thing that you can configure: entrypoints. The language server runs `elm make` to get type errors. By default `elm make` is run on the current file only. To get errors for the entire project you can specify your entrypoint files – basically, those with `main =` in them. Then the language server will run `elm make` on those instead.
```
lydell, line 93: Note 100% sure if this is true

Razzeee: The only slightly lieish thing would be `basically, those with main =`  I guess you could also argue, that `entire project` is technically wrong, but not sure how to better say it.

PR #319, opened 2020-09-07
Make declaration from usage

Razzeee:


Razzeee:
I think this is ready for a first version.

src/providers/codeActionProvider.ts
```diff
@@ -282,4 +287,62 @@ export class CodeActionProvider {
...
+          nodeAtPosition.tree,
+          params.textDocument.uri,
+          elmWorkspace.getImports(),
+        );
+
+        const edit = RefactorEditUtils.createTopLevelFunction(
+          insertLineNumber ?? tree.rootNode.endPosition.row,
+          funcName,
+          typeString,
+          nodeAtPosition.parent?.parent?.parent,
```
jmbockhorst, line 329: Using `TreeUtils.parentOfType('value_declaration', nodeAtPosition)` might be safer and cleaner.

src/providers/codeActionProvider.ts
```diff
@@ -282,4 +287,62 @@ export class CodeActionProvider {
...
+        );
+
+        if (edit) {
+          codeActions.push({
```
jmbockhorst, line 333: Should would try to put this code action above ones from elm make? Otherwise it can be easy to miss

jmbockhorst (approved):
Works good, just a few minor comments. Type inference doesn't work for functions with no args (I knew it wouldn't), so I will add that piece tonight in a separate PR.

src/providers/codeActionProvider.ts
Razzeee: Changed

src/providers/codeActionProvider.ts
Razzeee: I've moved these and the type annotation ones to the top now

PR #320, opened 2020-09-26
Parse files incrementally and use more type caching

jmbockhorst:
Work done to switch the tree-sitter parsing to incremental, so we can cache types between changes. I actual did most of this before I wrote inference. This part will need the most testing, but it seems to be working pretty good for me. 

I also added type inference record completions when all other records completions fail. This should be safe since it is only used as a backup for now (still a little slower than current methods), and won't break anything if it fails. This leads to some pretty cools results and it worked pretty well in my brief testing.

Things like:
``` elm
test: List Data -> Data
test data =
    data |> List.filter (\d -> d.{-caret-})
```
now work great!
I was kinda surprised at how much just worked.

I expect that this PR will probably need some cleanup and good testing before we want to merge.

### DIFF
--- a/src/providers/astProvider.ts
+++ b/src/providers/astProvider.ts
@@ -92,16 +92,15 @@ export class ASTProvider {
           : undefined,
       ])
       .forEach(([startNode, endNode]) => {
-        if (startNode || endNode) {
-          elmWorkspace
-            .getTypeCache()
-            .invalidateValueDeclaration(startNode ?? endNode!);
-
-          if (startNode && endNode) {
-            elmWorkspace.getTypeCache().invalidateValueDeclaration(endNode);
-          }
+        if (
+          startNode &&
+          endNode &&
+          startNode.id === endNode.id &&
+          TreeUtils.getTypeAnnotation(startNode)
+        ) {
+          elmWorkspace.getTypeCache().invalidateValueDeclaration(startNode);
         } else {
-          elmWorkspace.getTypeCache().invalidateAll();
+          elmWorkspace.getTypeCache().invalidateProject();
         }
       });
 
--- a/src/providers/astProvider.ts
+++ b/src/providers/astProvider.ts
@@ -9,15 +9,18 @@ import {
   Emitter,
 } from "vscode-languageserver";
 import { URI } from "vscode-uri";
-import Parser, { Tree, SyntaxNode } from "web-tree-sitter";
+import Parser, { Tree, Edit, Point } from "web-tree-sitter";
 import { IElmWorkspace } from "../elmWorkspace";
-import { IDocumentEvents } from "../util/documentEvents";
 import { ElmWorkspaceMatcher } from "../util/elmWorkspaceMatcher";
+import { Position, Range } from "vscode-languageserver-textdocument";
 import { FileEventsHandler } from "./handlers/fileEventsHandler";
+import { TextDocumentEvents } from "../util/textDocumentEvents";
+import { TreeUtils } from "../util/treeUtils";
 
 export class ASTProvider {
   private connection: IConnection;
   private parser: Parser;
+  private documentEvents: TextDocumentEvents;
 
   private treeChangeEvent = new Emitter<{ uri: string; tree: Tree }>();
   readonly onTreeChange: Event<{ uri: string; tree: Tree }> = this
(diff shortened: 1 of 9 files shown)


Razzeee:
As I want to build a version and post it to slack for extended testing, this probably won't hurt too much?

src/util/types/syntaxNodeMap.ts
```diff
@@ -52,4 +52,16 @@ export class SyntaxNodeMap<K extends SyntaxNode, V> {
       callback(val, ({ id: key } as unknown) as K),
     );
   }
+
+  public clear(): void {
+    this.map.clear();
+  }
+
+  public delete(key: K): void {
+    if (!("id" in key)) {
```
Razzeee, line 61: This looks, like it misses a typing?

src/util/types/typeCache.ts
```diff
@@ -36,6 +51,22 @@ export class TypeCache {
...
+      case "PROJECT_TYPE_ANNOTATION":
+        return this.projectTypeAnnotation.getOrSet(node, setter);
+      case "PROJECT_TYPE_AND_TYPE_ALIAS":
+        return this.projectTypeAndTypeAlias.getOrSet(node, setter);
+      case "PROJECT_VALUE":
+        return this.projectValue.getOrSet(node, setter);
     }
   }
+
+  public invalidateAll(): void {
```
Razzeee, line 63: This seems to only invalidate all project ones?

src/util/types/syntaxNodeMap.ts
jmbockhorst: Yeah, will update with the other PR merge.

src/util/types/typeCache.ts
```diff
@@ -5,12 +5,18 @@ import { InferenceResult } from "./typeInference";
...
+  | "PROJECT_TYPE_AND_TYPE_ALIAS"
+  | "PROJECT_VALUE";
 
 export class TypeCache {
   private packageTypeAnnotation: SyntaxNodeMap<SyntaxNode, InferenceResult>;
   private packageTypeAndTypeAlias: SyntaxNodeMap<SyntaxNode, InferenceResult>;
   private packageValue: SyntaxNodeMap<SyntaxNode, InferenceResult>;
+  private projectTypeAnnotation: SyntaxNodeMap<SyntaxNode, InferenceResult>;
+  private projectTypeAndTypeAlias: SyntaxNodeMap<SyntaxNode, InferenceResult>;
+  private projectValue: SyntaxNodeMap<SyntaxNode, InferenceResult>;
```
Razzeee, line 19: Would `projectValueDeclaration` be clearer here?

src/util/types/typeCache.ts
```diff
@@ -5,12 +5,18 @@ import { InferenceResult } from "./typeInference";
...
-  | "PACKAGE_VALUE";
+  | "PACKAGE_VALUE"
+  | "PROJECT_TYPE_ANNOTATION"
+  | "PROJECT_TYPE_AND_TYPE_ALIAS"
+  | "PROJECT_VALUE";
 
 export class TypeCache {
   private packageTypeAnnotation: SyntaxNodeMap<SyntaxNode, InferenceResult>;
   private packageTypeAndTypeAlias: SyntaxNodeMap<SyntaxNode, InferenceResult>;
   private packageValue: SyntaxNodeMap<SyntaxNode, InferenceResult>;
```
Razzeee, line 16: Would `packageValueDeclaration` be clearer here?

jmbockhorst:
I do need to clean up the cache invalidating a little tonight. But then it should be good to include with the build to test

src/util/types/typeCache.ts
jmbockhorst: Yep.

src/util/types/typeCache.ts
jmbockhorst: Yes. We should never need to invalidate package types, I think?

src/providers/completionProvider.ts
```diff
@@ -786,6 +801,32 @@ export class CompletionProvider {
     return result;
   }
 
+  private getRecordCompletionsUsingInference(
```
Razzeee, line 804: Might be worth to add a test for this?

jmbockhorst: Yes, will tonight.

src/providers/astProvider.ts
```diff
@@ -97,4 +131,62 @@ export class ASTProvider {
...
+      line: lines.length - 1,
+      character: lines[lines.length - 1].length,
+    };
+  }
+
+  private getIndexesFromRange(range: Range, text: string): [number, number] {
+    let startIndex = range.start.character;
+    let endIndex = range.end.character;
+
+    const regex = new RegExp(/\r\n|\r|\n/);
```
Razzeee, line 166: Condense this down like above?

src/util/types/typeCache.ts
Razzeee: Probably correct, but saying `All` here might be misleading? And we will probably want to refresh package types, as soon as we start to track those. It's pretty unexpected for newbies, that we don't pick them up after you've added something to your elm-json.

jmbockhorst:
Ready for merge.

PR #322, opened 2020-09-27
Implement elm-analyse like diagnostics natively using tree-sitter

jmbockhorst:
Had some fun using tree-sitter queries to start implementing some elm-analyse functionality. Currently, there is diagnostics for unsused import value and type, and unused import alias. Still working on more, just wanted to show the progress and how powerful using queries are. 

### Elm Analyse checks implemented:
- [x] Boolean Case Expression
- [x] Drop Concat Of Lists
- [x] Drop Cons Of Item And List
- [x] Fully Applied Operator as Prefix
- [x] Map Nothing To Nothing
- [x] No Top Level Signature
- [x] Single Field Record
- [x] Unnecessary List Concat
- [x] Unnecessary Port Module
- [x] Unused Import
- [x] Unused Import Alias
- [x] Unused Imported Variable
- [x] Unused Pattern Variable
- [x] Unused Top Level
- [x] Unused Type Alias
- [x] Unused Value Constructor
- [x] Unused Variable
- [x] Use Cons Over Concat


src/util/refactorEditUtils.ts
```diff
@@ -50,22 +50,29 @@ export class RefactorEditUtils {
       const exposedValues = TreeUtils.descendantsOfType(
```
Razzeee, line 50: Might want to rename this

jmbockhorst:
@Razzeee I added the list of checks that elm-analyse does and which ones I have implemented so far. Let me know or edit the list if there are some that you don't want to bring over. We can discuss, but I feel like there are some that we won't want to implement.

Razzeee:
Agreed, I would however not try to do all in one PR, so that list might be best moved to an issue.

.github/workflows/npm-publish.yml
```diff
@@ -1,47 +0,0 @@
-# This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
```
Razzeee, line 1: Probably a rebase error

jmbockhorst: Then again, I'm not quite sure.

jmbockhorst:
Ready for review. All of the tests from elm-analyse pass, but there could still be wrong cases out there. I left elm-analyse in place so that we can do some manual testing and compare the diagnostics. Once we feel it is good enough, we can disable these when elm-analyse is on, or just remove elm-analyse all together.

Of course we can now do much more than elm-analyse ever did, but as a start, I want to implement it a close to the existing behavior. 

Feel free to change/reword the diagnostics messages.

Razzeee:
Great work, I will hopefully find some time next week to look at this, this weekend is pretty stuffed :)

src/providers/diagnostics/fileDiagnostics.ts
```diff
@@ -6,6 +6,7 @@ export const enum DiagnosticKind {
   ElmAnalyse,
   ElmTest,
   TypeInference,
+  Elm,
```
Razzeee, line 9: I'm conflicted on this name, might make more sense to name/report this as ElmLS? What do you think?

Razzeee (changes requested):

src/providers/diagnostics/fileDiagnostics.ts
jmbockhorst: Yeah I didn't know what to name it. I'm fine with ElmLS.

src/providers/diagnostics/elmDiagnostics.ts
```diff
@@ -0,0 +1,960 @@
...
+  }
+
+  public createDiagnostics = (
+    tree: Tree,
+    uri: string,
+    elmWorkspace: IElmWorkspace,
+  ): Diagnostic[] => {
+    try {
+      return [
+        ...this.getUnusedImportDiagnostics(tree),
```
Razzeee, line 41: This is fine for now, but I think we will need at least the path ignore from here in the long run https://stil4m.github.io/elm-analyse/#/configuration

Razzeee:
Those changes, a rebase and a global on/off switch and we're good to go.

jmbockhorst:
Should be good now. Added a watch for the configuration so it updates dynamically.

test/diagnosticTests/elmDiagnostics.test.ts
```diff
@@ -12,7 +12,7 @@ import { baseUri } from "../utils/mockElmWorkspace";
 import { getSourceFiles } from "../utils/sourceParser";
```
Razzeee, line 12: This file should be renamed

PR #323, opened 2020-09-28
Fix inference bugs after feedback

jmbockhorst:
Fixing inference bugs after feedback.


jmbockhorst:
This should be ready go. All bugs that have been noted in Slack are fixed with this.

PR #324, opened 2020-09-30
Fix anonymous function param definitions and add code actions for let expr functions

jmbockhorst:
Fixed anonymous function param definitions and was able to enable some existing tests. This fixed some inference bugs. 
Also added a code action to add the type annotation to functions in let expr. 

Inference bugs fixed:
- WebGl type renders as `shader` (like intellij)
- Use module alias if there is one
- Fix functions that used anonymous functions with record or tuple param deconstructors (like `tuple` in `elm-codec`)

All other reported bugs behave the same as intellij-elm, so I don't think we should worry about those edge cases for now.

### DIFF
--- a/src/util/treeUtils.ts
+++ b/src/util/treeUtils.ts
@@ -1421,31 +1421,20 @@ export class TreeUtils {
     node: SyntaxNode,
     functionParameterName: string,
   ): SyntaxNode | undefined {
-    if (
-      node.type === "parenthesized_expr" ||
-      node.type === "anonymous_function_expr"
-    ) {
-      const anonymousFunctionExprNodes =
-        node.type === "anonymous_function_expr"
-          ? [node]
-          : this.descendantsOfType(node, "anonymous_function_expr");
-      const match = anonymousFunctionExprNodes
-        .map((a) => a.children)
-        .reduce((a, b) => a.concat(b), [])
-        .find(
-          (child) =>
-            child.type === "pattern" && child.text === functionParameterName,
-        )?.firstNamedChild;
+    const anonymousFunctionExprNodes = TreeUtils.getAllAncestorsOfType(
+      "anonymous_function_expr",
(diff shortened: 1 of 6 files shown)


test/definitionProviderTests/miscDefinition.test.ts
```diff
@@ -52,49 +52,50 @@ foo =
 
   // LAMBDAS (ANONYMOUS FUNCTIONS)
 
-  xit(`test lambda parameter ref`, async () => {
+  it(`test lambda parameter ref`, async () => {
     const source = `
 f = \\x -> x
    --X  --^
 `;
+    console.log(source);
```
Razzeee, line 60: Should we remove this log line?

test/definitionProviderTests/miscDefinition.test.ts
```diff
@@ -52,49 +52,50 @@ foo =
...
+  it(`test lambda parameter destructured tuple ref`, async () => {
     const source = `
-f = \(x,y) -> x
+f = \\(x,y) -> x
     --X     --^
 `;
     await testBase.testDefinition(source);
   });
 
   xit(`test lambda parameter destructured with alias`, async () => {
```
Razzeee, line 96: I guess this one isn't fixed yet?

src/util/treeUtils.ts
```diff
@@ -2156,6 +2145,11 @@ export class TreeUtils {
         return;
       }
 
+      const asClause = TreeUtils.findFirstNamedChildOfType(
```
Razzeee, line 2148: Only do this if it's needed. So after the next `return`

test/definitionProviderTests/miscDefinition.test.ts
jmbockhorst: Turns out it was fixed.

PR #326, opened 2020-10-02
Fix more definition and inference bugs

jmbockhorst:
Fix some bugs due to type vs union constructor definitions, if else expr parsing, and fix/add/enable more tests.

### DIFF
--- a/test/definitionProviderTests/importDefinition.test.ts
+++ b/test/definitionProviderTests/importDefinition.test.ts
@@ -11,6 +11,7 @@ import Foo exposing (bar)
 --@ Foo.elm
 module Foo exposing (bar)
 bar = 42
+--X
 `;
     await testBase.testDefinition(source);
   });
@@ -24,6 +25,7 @@ main = bar
 --@ Foo.elm
 module Foo exposing (bar)
 bar = 42
+--X
 `;
     await testBase.testDefinition(source);
   });
@@ -76,6 +78,7 @@ f = bar
 --@ Foo.elm
 module Foo exposing (bar)
 bar = 42
+--X
 `;
     await testBase.testDefinition(source);
   });
@@ -89,6 +92,7 @@ f = bar
 --@ Foo.elm
 module Foo exposing (..)
 bar = 42
+--X
 `;
     await testBase.testDefinition(source);
   });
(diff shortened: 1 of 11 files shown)


test/definitionProviderTests/importDefinition.test.ts
```diff
@@ -113,6 +114,7 @@ import App exposing (Page)
 --@ App.elm
 module App exposing (Page)
 type Page = Home
+--X
```
Razzeee, line 117: Are these changes necessary? So far the code only checks if we changed to the correct file I think, are you planning to change this?

test/definitionProviderTests/qualifiedImportDefinition.test.ts
```diff
@@ -105,6 +105,44 @@ update msg model = (model, Ports.foo "blah")
 --@ Ports.elm
 port module Ports exposing (foo)
 port foo : String -> Cmd msg
+`;
+    await testBase.testDefinition(source);
+  });
+
+  xit(`test union constructor should not resolve to type declaration from other file`, async () => {
```
Razzeee, line 112: Still deactivated?

test/definitionProviderTests/importDefinition.test.ts
jmbockhorst: I had added the ability to check the node position too in https://github.com/elm-tooling/elm-language-server/commit/f63617aa42fb0b40519187cfc065f6d5a1416bc4#.

test/definitionProviderTests/qualifiedImportDefinition.test.ts
jmbockhorst: Yeah, for some reason it wouldn't select the right node. It resolves to `User { data : String }` starting at character 12, which is correct. But the `--X` was selecting a node starting at character 10. Not sure why

jmbockhorst: Looks like I didn't hook it up right.

test/definitionProviderTests/importDefinition.test.ts
Razzeee: Right, forgot about that :)

PR #328, opened 2020-10-03
More fixes for union constructors, enabling more tests

jmbockhorst:
Fixes for union constructors and imports. One of the big problems with the tests is we were using `exposing (Page(Home))`, which isn't valid syntax anymore (as of 0.19). Intellij-elm has since fixed their tests and removed irrelevant ones, which I did as well. Was able to enable pretty much all of the type and union constructor tests, so I feel pretty good about it.

### DIFF
--- a/src/imports.ts
+++ b/src/imports.ts
@@ -288,19 +288,17 @@ import Platform.Sub as Sub exposing ( Sub )
       .forEach((element) => {
         if (element.exposedUnionConstructors) {
           result.push(
-            ...element.exposedUnionConstructors
-              .filter((a) => a.accessibleWithoutPrefix)
-              .map((a) => {
-                return {
-                  alias: `${a.name}`,
-                  fromModuleName: moduleName,
-                  fromUri: uri,
-                  maintainerAndPackageName,
-                  node: a.syntaxNode,
-                  type: "UnionConstructor" as NodeType,
-                  explicitlyExposed: false,
-                };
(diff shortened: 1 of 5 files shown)


src/imports.ts
```diff
@@ -334,16 +332,32 @@ import Platform.Sub as Sub exposing ( Sub )
...
+            alias: a.name,
+            fromModuleName: moduleNameNode.text,
+            fromUri: foundModule.uri,
+            maintainerAndPackageName: foundModule.maintainerAndPackageName,
+            node: a.syntaxNode,
+            type: a.type,
+            explicitlyExposed: true,
+          },
+        ].concat(
+          a.exposedUnionConstructors?.map((b) => {
```
jmbockhorst, line 348: We were missing these, so for an import like `import App exposing (Page(..))`, the union constructor for `Page` should be accessible without a prefix.

src/imports.ts
```diff
@@ -288,19 +288,17 @@ import Platform.Sub as Sub exposing ( Sub )
       .forEach((element) => {
         if (element.exposedUnionConstructors) {
           result.push(
-            ...element.exposedUnionConstructors
-              .filter((a) => a.accessibleWithoutPrefix)
```
jmbockhorst, line 292: We shouldn't filter here, because for `import App exposing (..)`, even the union constructors should be accessible without prefixes.

PR #335, opened 2020-10-06
Fix function hovers

Razzeee:
Not very happy with the needed adjustment for functions, but I guess it's the most reliable hotfix.

See https://github.com/elm-tooling/elm-language-client-vscode/issues/118

### DIFF
--- a/src/providers/hoverProvider.ts
+++ b/src/providers/hoverProvider.ts
@@ -42,7 +42,7 @@ export class HoverProvider {
         params.position,
       );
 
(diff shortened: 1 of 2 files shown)


jmbockhorst:
Yeah, I changed function definitions to go to the `function_declaration_left` because of the inference algorithm, but I'm not happy with that change either. I will revisit the type inference and try to revert to the old behavior, now that we have good tests.

Razzeee:
I'll move on with this as a bugfix.

Issue #337, opened 2020-10-07
language-server doesn't provide Content-Length header


romariolopezc:
This error just started happening using Neovim Coc.nvim. According to [this](https://github.com/neoclide/coc.nvim/issues/1327), this is not a bug in coc.nvim but in the language server. Its supposed that the language server implemens this header

I searched the specification https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/ and it says something about the content-length header

## Current Behavior

Uncaught exception logged in coc.nvim
![image](https://user-images.githubusercontent.com/6199588/95285534-a2eaac00-0826-11eb-9865-9794cfee08e7.png)

## Steps to Reproduce (for bugs)

1. Update to latest Coc.nvim
1. Install elm-language-server
1. Open an elm project

## Context

Is hard to write when this is popping up in each keystroke

## Your Environment

```
vim version: NVIM v0.5.0-590-gb39edb5b6
node version: v10.17.0
coc.nvim version: 0.0.79-b1288b0de4
coc.nvim directory: /home/rlopez/.local/share/nvim/plugged/coc.nvim
term: alacritty
platform: linux
```

Razzeee:
A new version fixing this was released.

PR #338, opened 2020-10-08
Catch exceptions from type inference

Razzeee:
We might want to do this in the type inference handlers instead. But this fix has been tested by our VIM users and has been confirmed resolving the problem.

https://github.com/elm-tooling/elm-language-server/issues/337

I think it might also make sense to open an issue in the language-server implementation lib? What's your POV @jmbockhorst

### DIFF
--- a/src/util/types/typeInference.ts
+++ b/src/util/types/typeInference.ts
@@ -42,6 +42,8 @@ import { Sequence } from "../sequence";
 import { Utils } from "../utils";
 import { RecordFieldReferenceTable } from "./recordFieldReferenceTable";
 import { TypeRenderer } from "./typeRenderer";
+import { container } from "tsyringe";
+import { IConnection } from "vscode-languageserver";
 
 export interface Alias {
   module: string;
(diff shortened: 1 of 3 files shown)


jmbockhorst:
It feels like it would be better to handle the exception in the `findType` function, so you only have to once, but this is fine.

src/util/types/typeInference.ts
```diff
@@ -2440,71 +2442,82 @@ export function findType(
...
-    return findTypeOrParentType(node) ?? TUnknown;
-  } else {
+      return findTypeOrParentType(node) ?? TUnknown;
+    } else {
+      return TUnknown;
+    }
+  } catch (error) {
+    const connection = container.resolve<IConnection>("Connection");
+    connection.console.warn(
+      `Error on type inference for type annotation codeAction. ${error}`,
```
jmbockhorst, line 2519: Should probably be a better description.

jmbockhorst (approved):

src/util/types/typeInference.ts
Razzeee: Whoops, that's actually a c&p error

PR #341, opened 2020-10-09
Use a better system for diagnostics with debouncing

jmbockhorst:
Use a better system for handling diagnostics. Only send a diagnostics message if the files diagnostics have changed and also debounce requests. Early testing seems like this improves performance by a lot in large workspaces.

### DIFF
--- a/src/providers/diagnostics/diagnosticsProvider.ts
+++ b/src/providers/diagnostics/diagnosticsProvider.ts
@@ -11,7 +11,9 @@ import { TextDocumentEvents } from "../../util/textDocumentEvents";
 import { ElmMakeDiagnostics } from "./elmMakeDiagnostics";
 import { TypeInferenceDiagnostics } from "./typeInferenceDiagnostics";
 import { ASTProvider } from "../astProvider";
-import { IElmWorkspace } from "src/elmWorkspace";
+import { IElmWorkspace } from "../../elmWorkspace";
+import { debounce } from "ts-debounce";
+import { DiagnosticKind, FileDiagnostics } from "./fileDiagnostics";
 
 export interface IElmIssueRegion {
   start: { line: number; column: number };
@@ -34,12 +36,7 @@ export class DiagnosticsProvider {
   private elmAnalyseDiagnostics: ElmAnalyseDiagnostics | null = null;
   private typeInferenceDiagnostics: TypeInferenceDiagnostics;
   private elmWorkspaceMatcher: ElmWorkspaceMatcher<{ uri: string }>;
-  private currentDiagnostics: {
-    elmMake: Map<string, Diagnostic[]>;
-    elmAnalyse: Map<string, Diagnostic[]>;
-    elmTest: Map<string, Diagnostic[]>;
-    typeInference: Map<string, Diagnostic[]>;
-  };
+  private currentDiagnostics: Map<string, FileDiagnostics>;
   private events: TextDocumentEvents;
   private connection: IConnection;
   private settings: Settings;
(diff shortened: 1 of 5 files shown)


src/providers/diagnostics/diagnosticsProvider.ts
```diff
@@ -28,18 +35,55 @@ export interface IElmIssue {
...
 }
 
+const enum DiagnosticKind {
+  ElmMake,
+  ElmAnalyse,
+  ElmTest,
+  TypeInference,
+}
+
+function rangeEquals(a: Range, b: Range): boolean {
```
Razzeee, line 45: I guess we should move this to a general place too, calling this should be helpful in other files.

src/providers/diagnostics/diagnosticsProvider.ts
```diff
@@ -140,75 +173,63 @@ export class DiagnosticsProvider {
...
-      }
+    if (fileDiagnostics) {
+      didUpdate = fileDiagnostics.update(kind, diagnostics);
+    } else if (diagnostics.length > 0) {
+      fileDiagnostics = new FileDiagnostics(uri);
+      fileDiagnostics.update(kind, diagnostics);
+      this.currentDiagnostics.set(uri, fileDiagnostics);
+      didUpdate = true;
     }
+    const sendDiagnostics = (uri: string): void => {
```
Razzeee, line 210: Do we really need to construct these when `didUpdate` is not true?

src/providers/diagnostics/diagnosticsProvider.ts
```diff
@@ -254,7 +284,37 @@ export class DiagnosticsProvider {
     ) {
       await this.elmAnalyseDiagnostics.updateFile(uri, text);
     }
+  }
+}
+
+class FileDiagnostics {
```
Razzeee, line 290: Move this to it's own file please :)

src/providers/diagnostics/diagnosticsProvider.ts
```diff
@@ -28,18 +35,55 @@ export interface IElmIssue {
...
+function rangeEquals(a: Range, b: Range): boolean {
+  return (
+    a.start.character === b.start.character &&
+    a.start.line === b.start.line &&
+    a.end.character === b.end.character &&
+    a.end.line === b.end.line
+  );
+}
+
+function diagnosticsEquals(a: Diagnostic, b: Diagnostic): boolean {
```
Razzeee, line 54: Should be moved out together with FileDiagnostics

src/providers/diagnostics/diagnosticsProvider.ts
```diff
@@ -254,7 +284,37 @@ export class DiagnosticsProvider {
...
+    return [
+      ...this.getForKind(DiagnosticKind.ElmMake),
+      ...this.getForKind(DiagnosticKind.ElmAnalyse),
+      ...this.getForKind(DiagnosticKind.ElmTest),
+      ...this.getForKind(DiagnosticKind.TypeInference),
+    ];
+  }
+
+  public update(kind: DiagnosticKind, diagnostics: Diagnostic[]): boolean {
+    const existing = this.diagnostics.get(kind);
```
Razzeee, line 308: Using `getForKind` might make sense, then you can drop the existing check in the next line

Razzeee (approved):
Code looks good to me, will have a runtime test and then merge.

jmbockhorst:
Cool. I also tested during work yesterday afternoon and didn't see anything unexpected.

PR #349, opened 2020-10-29
Fix some type inference bugs

jmbockhorst:
Found some inference bugs while working on using type inference to show compiler diagnostics.


Razzeee:
Seems to conflict, are there any test cases we should add?

Razzeee (changes requested):
Looks good otherwise

jmbockhorst:
Ready to merge.

src/util/treeUtils.ts
```diff
@@ -2190,7 +2211,7 @@ export class TreeUtils {
             child.firstNamedChild?.text,
         ),
       )
-      .filter(Utils.notUndefined.bind(this))
+      .filter(Utils.notUndefined)
```
Razzeee, line 2214: Is there a reason your removing this?

jmbockhorst: No, sorry probably happened in the merge.

Razzeee: Readded

Issue #352, opened 2020-10-30
Vscode: Notification handler 'textDocument/didOpen' failed with message: Cannot read property 'uri' of undefined


whitecolor:
There is an error that happens when opening elm file:
```
[Error - 10:22:43 PM] Notification handler 'textDocument/didOpen' failed with message: Cannot read property 'uri' of undefined
```

![image](https://user-images.githubusercontent.com/736697/97737339-acea8c80-1afe-11eb-8d32-c16bc9b6a184.png)

## Possible Solution

Here:

https://github.com/elm-tooling/elm-language-server/blob/87d4be08b9c5f029379f7d897f0935af70ba9825/src/providers/astProvider.ts#L48

and 

https://github.com/elm-tooling/elm-language-server/blob/87d4be08b9c5f029379f7d897f0935af70ba9825/src/providers/astProvider.ts#L58

`params` happen to be of another type do not contain `textDocument` but `document` field with another structure.

## Context
Latest VScode and Insiders.

Augustin82:
Might be the same problem as https://github.com/elm-tooling/elm-language-server/issues/348 ? However, this wouldn't explain why it seems to occur "randomly", right?

whitecolor:
Well no, for me in the simplest setup it happens always as new file opens.

Razzeee:
That's correct, see the merge request I attached

Issue #348, opened 2020-10-26
No errors visible in neovim + coc.


brzezinskip:
## Expected Behavior
ELS 1.13.1 + COC combination should display errors and warnings in neovim (it does in VSC). It does work fine on 1.11.0.

## Current Behavior
No errors are being displayed. CocInfo, amongst other lines, has this one `[Error  - 10:12:16] Notification handler 'textDocument/didOpen' failed with message: Cannot read property 'uri' of undefined` which I believe may be the root of the issue here.

## Possible Solution
Unfortunately, no idea. I've been trying to find out what library could be causing the above error but no luck.

## Steps to Reproduce (for bugs)
1. Install 1.13.1 version of ELS.
2. Open any .elm file
3. Make a mistake so that compiler fails.
4. Notice there's no error shown in the place you broke the code at.

## Context
I've tried upgrading my ELS because of some other small issue I've been having in older versions. I did that by cloning the repo and manually linking. Unfortunately it didn't work, so I unlinked and installed the latest version through npm. No luck there either, so in the end, I just reverted to 1.11.0 as it worked fine there before for me

## Your Environment
* Version used: 1.13.1
* Editor used: Neovim
* Environment name and version (e.g. node.js 5.4): node v12.13.1
* Operating System and version: osx catalina 10.15.7

Razzeee:
This is the relevant version diff https://github.com/elm-tooling/elm-language-server/compare/1.11.0...1.12.0

andys8:
I did give it a try with my setup. Errors seem to work fine with `elm-language-server` 1.13.1.

```
vim version: VIM - Vi IMproved 8.2 8021704
node version: v12.15.0
coc.nvim version: 0.0.79-d3293597d1
```

Can you reproduce it with a public repository, too?
What's the exact version of CoC you're using?

brzezinskip:
Hey Andy, here's an info I've gotten from CocInfo

```
vim version: NVIM v0.4.4
node version: v12.13.1
coc.nvim version: 0.0.79-230b8a7247
```
I'll try reproducing with a public repo asap. Thanks for looking into it!

Augustin82:
I've got this on one computer. It works fine on the other, and I cannot figure out a difference in setups (apart from the problematic computer running node 14, but I since switched to node 10 and it didn't fix anything).

```
[Trace - 11:15:22 PM] Sending notification 'textDocument/didOpen'.
Params: {
    "textDocument": {
[...]

PR #353, opened 2020-10-31
Always return params from textDocumentEvents

Razzeee:
Closes https://github.com/elm-tooling/elm-language-server/issues/352 and https://github.com/elm-tooling/elm-language-server/issues/348


jmbockhorst (approved):
Looks good.

PR #354, opened 2020-10-31
Load imports lazily

### LINKED ISSUE #350: Parse file changes lazily
We shouldn't be parsing the files on every change, only when we need them.

jmbockhorst:
Closes #350.


jmbockhorst:
@Razzeee I am going to be testing this heavily over the next few days, but it is ready to start reviewing. Changed a lot to improve performance:
- Store tree containers using a `Map` for faster access. 
- Introduce a `TypeChecker` interface (inspired by Typescript) that handles type inference, definitions, and import resolves. 
- Do things lazily as much as possible, which saves a lot of unneeded processing between many changes.

jmbockhorst:
I had a project that was taking 60s to infer an entire workspace and found that 80% of the time was spent resolving definitions. With these changes I've been able to get that time down to 6s. Still work to be done, but I think this will help a lot

Big changes for performance optimizations:
- Introduce a binder that binds all variables to the containing scope. This makes definition resolves fast because it just has to traverse up the parent scopes and find a match in the symbol map.
- Don't use Object.assign in the expression tree mapper. I think this is a improvement but I still want to compare how much.

I'm also working on a consistent way to do performance testing on a workspace.

test/definitionProviderTests/typeResolveDefinition.test.ts
```diff
@@ -174,26 +174,6 @@ User { data = "" }
   type UnitlessFloat
   = UnitlessFloat
     --X   --^
-`;
-    await testBase.testDefinition(source);
-  });
-
-  it(`test union contructor should not resolve to type alias`, async () => {
```
Razzeee, line 181: Is this meant to be removed?

Razzeee:
Reminder to push the tree-sitter changes

test/definitionProviderTests/typeResolveDefinition.test.ts
jmbockhorst: Yes. It's actually not valid in the compiler to have two constructors (Union and record) with the same name in one file

src/elmWorkspace.ts
```diff
@@ -55,8 +62,8 @@ export class ElmWorkspace implements IElmWorkspace {
       `Starting language server for folder: ${this.rootPath.toString()}`,
     );
 
-    this.imports = new Imports();
     this.typeCache = new TypeCache();
+    this.possibleImportsCache = new PossibleImportsCache();
```
Razzeee, line 66: Can we move this to the dependency injection? (Maybe typeCache too)

src/providers/completionProvider.ts
```diff
@@ -44,6 +43,33 @@ interface ICompletionOptions {
   filterText?: string;
 }
 
+export interface IPossibleImportsCache {
```
Razzeee, line 46: Move this to it's own file

src/providers/completionProvider.ts
```diff
@@ -44,6 +43,33 @@ interface ICompletionOptions {
...
+  get(uri: string): IPossibleImport[] | undefined;
+  set(uri: string, possibleImports: IPossibleImport[]): void;
+  clear(): void;
+}
+
+export class PossibleImportsCache implements IPossibleImportsCache {
```
Razzeee, line 52: Move this to it's own file

src/elmWorkspace.ts
```diff
@@ -149,6 +177,7 @@ export class ElmWorkspace implements IElmWorkspace {
         type === "application"
           ? {
               ...elmJson.dependencies.direct,
+              ...elmJson.dependencies.indirect,
```
Razzeee, line 180: We should test this, I remember this being kind of hard to do (we probably don't have anything in the test suite to test this)

Razzeee:
It's also still conflicting FYI

jmbockhorst:
Made some more optimizations and fixed the conflicts. It's now down to about 4s to infer an entire workspace, down from 60s when I started.

jmbockhorst:
This is feeling pretty stable now and ready to merge when you are ready. I continue to find places to make performance optimizations and will be heavily testing this at work next work.

src/elmWorkspace.ts
```diff
@@ -111,6 +152,22 @@ export class ElmWorkspace implements IElmWorkspace {
         `Could not figure out elm version, this will impact how good the server works. \n ${e.stack}`,
       );
     }
+
+    // const options = {
```
Razzeee, line 156: Can this be removed?

src/util/treeUtils.ts
```diff
@@ -1404,114 +446,9 @@ export class TreeUtils {
...
-          if (foundNode) {
-            return foundNode;
-          }
-        }
-      }
-    }
-  }
-
+  /**
+   * @deprecated Should not use do to performance. Use bindings instead
```
Razzeee, line 450:
```suggestion
   * @deprecated Should not be used due to performance. Use bindings instead
```

src/util/treeUtils.ts
```diff
@@ -2151,104 +988,9 @@ export class TreeUtils {
...
-          exposedUnionConstructors: undefined,
-          name: portNode.children[1].text,
-          syntaxNode: portNode,
-          type: "Port",
-        };
-      });
-  }
-
+  /**
+   * @deprecated Should not use do to performance. Use bindings instead
```
Razzeee, line 992:
```suggestion
   * @deprecated Should not be used due to performance. Use bindings instead
```

src/util/types/binder.ts
```diff
@@ -0,0 +1,417 @@
...
+      });
+    } else {
+      // If there is a pattern, bind it to the parent container
+      const pattern = node.childForFieldName("pattern");
+
+      if (pattern) {
+        pattern.descendantsOfType("lower_pattern").forEach((lowerPattern) => {
+          container.set(lowerPattern.text, {
+            node: lowerPattern,
+            type: "Function", // This isn't a good type
```
Razzeee, line 123: This might need more comment? I don't think I understand it :)

jmbockhorst: It's a pattern inside a let declaration:
```elm

...
let
    { pattern } = ...

in
...
```

And we don't really have a good type for it right now.

Issue #357, opened 2020-11-03
Stale Errors


bengolds:
I often have the last few errors remaining in a file appearing _after_ I've fixed them. It seems to happen often if there are errors in other files, but I also have instances where the whole project is compiling and errors still haven't cleared. I don't have a ton of consistent, clean repros yet, but I'll add more as I get them.

## Expected Behavior
When all errors are fixed in a file, LSP should report 0 errors.

## Current Behavior
Often, after clearing all the errors out of a file, the last few remain and show old error messages.

## Steps to Reproduce (for bugs)
https://cln.sh/c6JA8C

## Your Environment
* Version used: Master, commit: e3c58f57a15106d0414ba881471568b47196c540
* Editor used: Neovim, CoC

PR #358, opened 2020-11-03
Reset diagnostics

Razzeee:
Fixes https://github.com/elm-tooling/elm-language-server/issues/357

### DIFF
--- a/src/providers/diagnostics/diagnosticsProvider.ts
+++ b/src/providers/diagnostics/diagnosticsProvider.ts
@@ -138,6 +138,8 @@ export class DiagnosticsProvider {
   private newElmAnalyseDiagnostics(
     diagnostics: Map<string, Diagnostic[]>,
   ): void {
+    this.resetDiagnostics(diagnostics, DiagnosticKind.ElmAnalyse);
+
     diagnostics.forEach((diagnostics, uri) => {
       this.updateDiagnostics(uri, DiagnosticKind.ElmAnalyse, diagnostics);
     });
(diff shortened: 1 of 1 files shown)


src/providers/diagnostics/diagnosticsProvider.ts
```diff
@@ -200,11 +200,13 @@ export class DiagnosticsProvider {
     const text = document.getText();
 
     if (isSaveOrOpen) {
-      const elmMakeDiagnostics = await this.elmMakeDiagnostics.createDiagnostics(
+      const diagnosticList = await this.elmMakeDiagnostics.createDiagnostics(
```
Razzeee, line 203: Didn't mean to change this

src/providers/diagnostics/diagnosticsProvider.ts
```diff
@@ -200,11 +200,13 @@ export class DiagnosticsProvider {
...
         uri,
       );
 
-      elmMakeDiagnostics.forEach((diagnostics, diagnosticsUri) => {
+      this.resetDiagnosticsForType(diagnosticList, DiagnosticKind.ElmMake);
```
Razzeee, line 207: Might be nicer to just push `[]` to elmMakeDiagnostics everything else should work fine then.

jmbockhorst:
Yeah this seems probably fine for now. Would have some problems with multiple entry points but that's pretty complicated.

Razzeee:
Added this for elm-analyse and elm-make now. I think I'm happy with the code, not sure how to add it for type inference diagnostics/if it needs it at all.

We need to test this a bit.

Razzeee:
My tests seemed successful. I only tested elm-make and elm-analyse.

@bengolds can you run this branch and test it for a while too?

bengolds:
So far, this is MILES better. No more stale errors yet!

PR #374, opened 2020-11-24
Fix union pattern when arg is a unit expr

jmbockhorst:
Fixes the elm-spa-example.

### DIFF
--- a/test/definitionProviderTests/importDefinition.test.ts
+++ b/test/definitionProviderTests/importDefinition.test.ts
@@ -475,4 +475,25 @@ quux = 99
 `;
     await testBase.testDefinition(source);
   });
+
+  it(`test union constructor resolves to the exposed one`, async () => {
+    const source = `
+--@ main.elm
+import App exposing (Page)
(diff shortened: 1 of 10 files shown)


test/definitionProviderTests/importDefinition.test.ts
```diff
@@ -472,6 +472,27 @@ bar = 42
...
 quux = 99
 --X
+`;
+    await testBase.testDefinition(source);
+  });
+
+  it(`test union constructor resolves to the exposed one`, async () => {
+    const source = `
+--@ main.elm
+import App exposing (Page)
```
Razzeee, line 482: I might be confused, but this isn't valid elm anymore, wasn't this changed? The test might still be valid.

jmbockhorst: As far as I’m aware, it is still valid. `exposing (Page(Home))` is what became invalid.

jmbockhorst:
Hopefully it should now be much more stable and any errors will be caught. I also did some work to improve record completions and show parsing diagnostics.

PR #378, opened 2020-11-25
Various fixes

### LINKED ISSUE #367: References not working anymore
Testing `master`, references are no longer working for:
- Imported qualified values like `Form.value`. Unqualified imports are still working
- Function parameters

We need to start adding tests for references.

### LINKED ISSUE #375: False unused imported value diagnostics
```elm
init json =
    case Json.Decode.decodeValue Json.flagsDecoder json of
        Ok flags ->
            ( Ready flags
                ({ empty
                    | ...
                 }
```

`empty` is falsely detected as being unused.

jmbockhorst:
- Fixes #375 
- Fix definition of `Svg.g` 
- Fix function not resolving to itself (affects references)
- Fix references when used with a module alias and start adding reference tests. Fixes #367

We need many more reference tests, but now the infrastructure is in place.


test/utils/sourceParser.ts
```diff
@@ -107,6 +111,14 @@ export function getInvokeAndTargetPositionFromSource(source: string): TestType {
           line: line - 1,
           character: targetCharacter + 2,
         };
+      } else {
+        targetCharacter = s.search(/(X)--/);
```
Razzeee, line 115: If I'm reading this right, I would prefer the pattern to be `--<-` as it seems what tree-sitter is moving to and I think intellij-elm is using that too?

Razzeee: https://github.com/tree-sitter/tree-sitter/pull/775/files#diff-90f214d3ccf2507a9f8b9224f93e9907d726b84bd32c00008f814039d74f1159R56

Razzeee (changes requested):
Looks great otherwise

test/utils/sourceParser.ts
jmbockhorst: Ok, I just did it so that I could make the target node be the first character.

Razzeee: Yeah, I got that, but let's use the other syntax, so we can keep it consistent.

jmbockhorst: I did `--<X` for now, feel free to change.

Razzeee (approved):

Issue #362, opened 2020-11-11
Type inference gets confused by aliases


miniBill:
Consider these two modules:

```
module LocalizedHtml exposing (Html, Language(..), div, text)

import Html


type Language
    = Italian
    | English


type alias Html a =
    Language -> Html.Html a


type alias Attribute a =
    Language -> Html.Attribute a


div : List (Attribute msg) -> List (Html msg) -> Html msg
div attributes children language =
    Html.div (List.map (\a -> a language) attributes) (List.map (\c -> c language) children)


text : String -> String -> Html msg
text en it language =
    Html.text <|
        case language of
            English ->
                en

            Italian ->
                it

```

and

```
module Main exposing (main)

import LocalizedHtml exposing (Html, Language(..), div, text)


main =
    div [] [ hello, world ] English


hello : Html msg
hello =
    text "Hello" "Ciao"


world =
    text "World" "Mondo"
```

notice how `Main` does *not* import `Html`.

## Expected Behavior
The type of `world` is inferred to be `Html msg`.

## Current Behavior
The type of `world` is inferred to be `Language -> Html msg`.

## Context
I'm wrapping `elm-ui` to add translations.

Razzeee:
@jmbockhorst can this be closed?

jmbockhorst:
No, I had to "unfix" this. The type inference is technically correct in this case, but it can maybe be rendered different because of the aliases. It's not a simple solution without adding more to the inference algorithm.

PR #379, opened 2020-11-27
More various fixes

jmbockhorst:
Also add script to run diagnostics on many elm packages. It's a little messy right now, but you can see which libs I had to skip due to parsing errors. Maybe we can add an action that runs nightly to detect new errors.


src/util/types/typeChecker.ts
```diff
@@ -402,11 +419,24 @@ export function createTypeChecker(workspace: IElmWorkspace): TypeChecker {
...
-          node: definitionNode.node,
-          nodeType: definitionNode.type,
-          uri,
-        };
+        let type: Type = TUnknown;
+        if (isConstructorUsage && isFunctionUsage) {
+          type = findType(definitionNode.node);
+        }
+
+        // If it is a constructor function usage, the type of the definition must be a function or record
```
Razzeee, line 427: I might be missing some context as I am on mobile, but what about ()?

src/util/types/typeInference.ts
```diff
@@ -1815,9 +1785,14 @@ export class InferenceScope {
     );
 
     if (variant?.expr.nodeType !== "UnionVariant") {
-      throw new Error(
-        "Could not find UnionVariant for " + unionPattern.constructor.text,
+      this.diagnostics.push(
+        error(
+          unionPattern,
+          Diagnostics.MissingValue,
+          unionPattern.constructor.lastNamedChild?.text ?? "union variant",
```
Razzeee, line 1792: We might want to look at the naming, not sure if it's the easiest to understand like this "union variant"

test/typeInference.test.ts
```diff
@@ -567,7 +567,7 @@ func =
     await testTypeInference(basicsSources + source, "String");
   });
 
-  test("function return with aliased type", async () => {
+  xtest("function return with aliased type", async () => {
```
Razzeee, line 570: Should this be commented out?

test/diagnostics.ts
```diff
@@ -0,0 +1,194 @@
...
+    }
+  } catch (e) {
+    console.log("\n");
+    console.log(uri);
+    console.log(e);
+    process.exitCode = 1;
+  }
+}
+
+function shuffleArray(arr: unknown[]): void {
```
Razzeee, line 111: Why shuffle? Wouldn't this just make the result inconsistent?

script/parse-examples-full
```diff
@@ -0,0 +1,61 @@
...
+examples_to_parse=$(
+  for example in $all_examples; do
+    echo $example
+  done
+)
+
+echo $examples_to_parse
+
+start=`date +%s.%N`
+tree_sitter_report=$(echo $examples_to_parse | npx ts-node ./test/diagnostics.ts)
```
Razzeee, line 43: You might want to rename this

script/parse-examples-full
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
```
Razzeee, line 1: There's no pipeline for this, should this even exist, as you seem to have moved most of this to code?

Razzeee (changes requested):

Razzeee:
My gut is also telling me that we might really want to look at that benchmarking lib

src/util/types/typeInference.ts
jmbockhorst: That case should never happen because the `lastNamedChild` should never be null, I just added it to make TS happy.

test/typeInference.test.ts
jmbockhorst: I actually had to un-fix #362 as the fix I made for it was wrong and broke other things. That issue is about type rendering, not inference. In that case, the type given is the proper type, but when rendered it should use the alias. However it's not so easy because the alias information gets lost along the way. So I will need to think more about how to properly fix that one.

test/diagnostics.ts
jmbockhorst: It just shuffles the libs that get run, but shouldn't make an inconsistent result, it will still run them all. I added this because when I was testing locally, I would only test 50 at a time, and I wanted to get a different sample each time. This is also why I used the `complete.txt` to save which ones were good between runs.

jmbockhorst:
Yeah I think so, but not for this script. This is just to make sure our type checking works like the compiler across all libraries.

script/parse-examples-full
jmbockhorst: Yeah, it should have been removed. I tried to use it at first, but couldn't get anything to work, so just did it in node.

src/util/types/typeChecker.ts
jmbockhorst: I don't think thats related. Basically this checks that if a constructor is used:

```elm
Model ""
```

We have to make sure the type of `Model` is a function or record. This failed if there was a local `Model` type alias defined, which may or may not actually be a record type. In one case it wasn't, and needed to use the imported one that was a union constructor.

test/diagnostics.ts
Razzeee: I'm strongly against shuffling in the CI, as it will make it hard to compare runs with another. (it's probably still hard enough)

jmbockhorst: Yeah, I’m fine to remove it.

PR #380, opened 2020-11-28
Add import resolve diagnostics and clean up API

jmbockhorst:
Some major refactoring of the type checker api for better handling diagnostics with an async option and allowing us to calculate other diagnostics like import resolves.

### DIFF
--- a/src/cancellation.ts
+++ b/src/cancellation.ts
@@ -187,6 +187,21 @@ export interface ICancellationToken {
   throwIfCancellationRequested(): void;
 }
 
+export class ServerCancellationToken implements ICancellationToken {
+  constructor(private cancellationToken: CancellationToken) {}
+
+  public isCancellationRequested(): boolean {
+    return this.cancellationToken.isCancellationRequested;
+  }
+
+  public throwIfCancellationRequested(): void {
+    if (this.isCancellationRequested()) {
+      console.log("throwing cancellation");
+      throw new OperationCanceledException();
+    }
+  }
+}
+
 /**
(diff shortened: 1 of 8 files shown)


src/cancellation.ts
```diff
@@ -187,6 +187,21 @@ export interface ICancellationToken {
...
+export class ServerCancellationToken implements ICancellationToken {
+  constructor(private cancellationToken: CancellationToken) {}
+
+  public isCancellationRequested(): boolean {
+    return this.cancellationToken.isCancellationRequested;
+  }
+
+  public throwIfCancellationRequested(): void {
+    if (this.isCancellationRequested()) {
+      console.log("throwing cancellation");
```
Razzeee, line 199: Should this be a console log? I thought we should use the connection to log?

src/util/types/typeChecker.ts
```diff
@@ -37,6 +38,25 @@ export interface DefinitionResult {
   nodeType: NodeType;
 }
 
+class DiagnosticsCollection extends Map<string, Diagnostic[]> {
```
Razzeee, line 41: I guess `ElmDiagnosticsHelper` should be merged with this? (not in this PR)

src/cancellation.ts
jmbockhorst: Meant to remove that.

src/util/types/typeChecker.ts
jmbockhorst: Yeah and we'll need to talk about the role of elm make (on save, never, controlled by a setting, etc) longer term.

PR #381, opened 2020-11-30
Fix import code actions

jmbockhorst:
Adds a new system for code actions using a registration model that should be nicer and easy to handle "fix all" code actions. Moves the import code action out of the `elmMakeDiagnostics` and into a shared code action registration that is used for type inference diagnostics as well. In the future I will work to consolidate other code actions to this new way. Also started adding tests for code actions. I assume it will need some cleanup, but its most of the way there. 

Another thing I changed was the diagnostics behavior between the compiler and our analyzer. Now, when you save, we clear our diagnostics and get them only from the compiler. Then if you make a change, we clear the compiler ones and only get ours. We can talk about this more if we need to.

### DIFF
--- a/src/providers/codeActionProvider.ts
+++ b/src/providers/codeActionProvider.ts
@@ -10,20 +10,36 @@ import {
 import { URI } from "vscode-uri";
 import { SyntaxNode, Tree } from "web-tree-sitter";
 import { IElmWorkspace } from "../elmWorkspace";
-import { ElmWorkspaceMatcher } from "../util/elmWorkspaceMatcher";
+import { ElmWorkspaceMatcher, IParams } from "../util/elmWorkspaceMatcher";
+import { MultiMap } from "../util/multiMap";
 import { RefactorEditUtils } from "../util/refactorEditUtils";
 import { Settings } from "../util/settings";
-import { TreeUtils } from "../util/treeUtils";
+import { flatMap, TreeUtils } from "../util/treeUtils";
+import { Diagnostics } from "../util/types/diagnostics";
 import { ElmLsDiagnostics } from "./diagnostics/elmLsDiagnostics";
 import { ElmMakeDiagnostics } from "./diagnostics/elmMakeDiagnostics";
+import { diagnosticsEquals } from "./diagnostics/fileDiagnostics";
 import { ExposeUnexposeHandler } from "./handlers/exposeUnexposeHandler";
 import { MoveRefactoringHandler } from "./handlers/moveRefactoringHandler";
+
+export type ICodeActionParams = CodeActionParams & IParams;
+
+export interface ICodeActionRegistration {
+  errorCodes: string[];
+  getCodeActions(params: ICodeActionParams): CodeAction[] | undefined;
+  getFixAllCodeAction(params: ICodeActionParams): CodeAction | undefined;
+}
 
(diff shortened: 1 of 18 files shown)


src/util/elmWorkspaceMatcher.ts
```diff
@@ -16,19 +22,40 @@ export class ElmWorkspaceMatcher<ParamType> {
     this.elmWorkspaces = container.resolve("ElmWorkspaces");
   }
 
+  public handle<ResultType>(
```
Razzeee, line 25: Is this the new default? Should we deprecate `handlerForWorkspace` and change the docs above?

src/providers/codeActionProvider.ts
```diff
@@ -45,37 +61,144 @@ export class CodeActionProvider {
...
-  ): CodeAction[] {
+  public static registerCodeAction(
+    registration: ICodeActionRegistration,
+  ): void {
+    registration.errorCodes.forEach((code) => {
+      CodeActionProvider.errorCodeToRegistrationMap.set(code, registration);
+    });
+  }
+
+  public static forEachDiagnostic(
```
Razzeee, line 72: private

src/providers/codeActionProvider.ts
```diff
@@ -45,37 +61,144 @@ export class CodeActionProvider {
...
+    params: ICodeActionParams,
+    title: string,
+    edits: TextEdit[],
+  ): CodeAction {
+    const changes = { [params.sourceFile.uri]: edits };
+    return {
+      title,
+      kind: CodeActionKind.QuickFix,
+      edit: { changes },
+      isPreferred: true,
```
Razzeee, line 97: I think we need a follow up ticket to handle this better. Just looked at the docs again and we're probably misusing it.

src/util/elmWorkspaceMatcher.ts
jmbockhorst: Yes we could deprecate, I wanted it to be opt-in for each handler because in some places we pass the params back to the client, and that's when it throws a json error (trying to stringify an `ElmWorkspace` ).

src/providers/codeActionProvider.ts
jmbockhorst: Yeah, I was unsure. I just left it how it was but it doesn't seem right to mark them all as preferred.

src/providers/codeActionProvider.ts
```diff
@@ -45,37 +61,144 @@ export class CodeActionProvider {
...
+
+    const diagnostics: Diagnostic[] = [];
+    CodeActionProvider.forEachDiagnostic(params, errorCodes, (diagnostic) => {
+      diagnostics.push(diagnostic);
+      callback(edits, diagnostic);
+    });
+
+    return {
+      title,
+      kind: CodeActionKind.QuickFix,
```
Razzeee, line 120: CodeActionKind.SourceFixAll seems to be what we sholud be doing

src/providers/codeActionProvider.ts
Razzeee: I think we might even want this to be false in this case (but need it configurable anyway)

Razzeee: Configurable as in this method needs to handle both :)

PR #382, opened 2020-12-01
Remove use of Diagnostic code property

jmbockhorst:
Adds a new interface `IDiagnostic` for the server that extends `Diagnostic` and ensures the `data` property has a `uri` and `code` and that we always omit the top level `code`. Also defined a new `Diagnostic` interface for the analyzer API so it doesn't rely on LSP.

### DIFF
--- a/src/providers/diagnostics/elmDiagnosticsHelper.ts
+++ b/src/providers/diagnostics/elmDiagnosticsHelper.ts
@@ -1,15 +1,15 @@
 import path from "path";
-import { Diagnostic, DiagnosticSeverity, Range } from "vscode-languageserver";
+import { DiagnosticSeverity, Range } from "vscode-languageserver";
 import { URI } from "vscode-uri";
 import { Diagnostics } from "../../util/types/diagnostics";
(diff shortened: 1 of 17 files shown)


src/providers/diagnostics/elmDiagnosticsHelper.ts
```diff
@@ -52,18 +52,18 @@ export class ElmDiagnosticsHelper {
...
-      `${messagePrefix}${issue.details.replace(/\[\d+m/g, "")}`,
-      this.severityStringToDiagnosticSeverity(issue.type),
-      code,
-      "Elm",
-    );
+    return {
+      range: lineRange,
+      message: `${messagePrefix}${issue.details.replace(/\[\d+m/g, "")}`,
+      severity: this.severityStringToDiagnosticSeverity(issue.type),
+      source: "elm",
```
Razzeee, line 65: Should be `Elm`

Razzeee (approved):
Seems like a nice refactoring, thanks!

PR #383, opened 2020-12-01
Refactor make declaration from usage code action

### LINKED ISSUE #377: "Create local function" action doesn't show up when asking code actions on a line
Coc in Neovim has three functions for asking for code actions: 
`coc-codeaction-selected` -- this fetches code actions for a selected area (a character, a word, a paragraph)
`coc-codeaction-line` -- this fetches code actions for a given line
`coc-codeaction` -- this fetches code actions for the whole file

`coc-codeaction-line` and `coc-codeaction` seem to show most code actions, with the exception of "create local function":
Screenshot 2020-11-25 at 09 24 32@2x
(this is `coc-codeaction`, or actions for the whole file)

`coc-codeaction-selected` on _just_ the function that was wrong shows the "create local function" code action.
Screenshot 2020-11-25 at 09 25 14@2x

From [this](https://gitter.im/neoclide/coc.nvim?at=5fbe19286a738b71d3612cd8) conversation in the Coc gitter, it appears as though the only difference is in the ranges that the LSP sends to the language server. So, `elm-language-server` may only be sending that action back for very specific ranges.

jmbockhorst:
Should fix #377 if it wasn't already fixed.


src/providers/codeAction/makeDeclarationFromUsageCodeAction.ts
```diff
@@ -0,0 +1,82 @@
...
+          params,
+          "Create local function",
+          edits,
+        ),
+      ];
+    }
+
+    return [];
+  },
+  getFixAllCodeAction: (params: ICodeActionParams): CodeAction | undefined => {
```
Razzeee, line 30: Just a heads up, I've seen this feature have a bug in the sense of it being offered, when it shouldn't be. Will have to track that down somewhen, but having a fix all might make it more likely to run into it.

src/util/utils.ts
```diff
@@ -50,4 +52,46 @@ export class Utils {
...
+    const result: T[] = [];
+    for (const item of array) {
+      if (!result.find((other) => equalityComparer(item, other))) {
+        result.push(item);
+      }
+    }
+    return result;
+  }
+
+  public static deduplicateDiagnostics(
```
Razzeee, line 77: This and the next one seem like the should be moved to diagnostics.

Razzeee:
Do you know, why exactly we need to deduplicate the diagnostics, is that by design?

jmbockhorst:
It was but I just thought about it more and we shouldn’t need to.

src/util/types/typeInference.ts
```diff
@@ -1041,14 +1041,19 @@ export class InferenceScope {
...
         ) as ETypeAnnotation,
         referenceUri,
         this.elmWorkspace,
         false,
-      )?.type;
+      );
+
+      if (result) {
+        type = result.type;
+        this.diagnostics.push(...result.diagnostics);
```
jmbockhorst, line 1055: Just need to get rid of this change, which is where to duplicates are coming from. I thought I needed this earlier but was thinking about it wrong.

src/providers/codeAction/makeDeclarationFromUsageCodeAction.ts
jmbockhorst: I noticed a function that got created with the wrong number of args. It was something like:

```elm
func : Model -> String -> Model
fund arg1 =
    ...
```

We should probably use the inferred annotation as the source to determine the number of args, as it will always be correct.

jmbockhorst:
Another thing to note is the fix all doesn't work for type annotation fixes yet because we use the program diagnostics to compute a fix all, but the type annotation diagnostics come from the provider, not the program/checker. That's a reason I was asking if we should move those down to the type checker.

jmbockhorst:
Refactored a bit more to add "suggestion" diagnostics into the type checker. This means that missing type annotation diagnostic comes from the type checker and "Fix all" work for adding them. Also split normal diagnostics into "semantic" and "syntactic" (Typescript does this) to get more control over them and to always get syntax errors first.

src/util/types/typeChecker.ts
```diff
@@ -869,4 +935,35 @@ export function createTypeChecker(workspace: IElmWorkspace): TypeChecker {
...
+    ).diagnostics.forEach((diagnostic) => diagnostics.add(diagnostic));
+  }
+
+  function checkTypeDeclaration(typeDeclaration: SyntaxNode): void {
+    TypeExpression.typeDeclarationInference(
+      mapSyntaxNodeToExpression(typeDeclaration) as ETypeDeclaration,
+      workspace,
+    ).diagnostics.forEach((diagnostic) => diagnostics.add(diagnostic));
+
+    // Need to check union variants
```
Razzeee, line 952: Is this a todo?

jmbockhorst: Nope.

PR #385, opened 2020-12-03
Add github action to test diagnostics

jmbockhorst:
This should give a "Run Manually" button on the actions tab. We could also set it up on a 24 hour schedule.

### DIFF
--- a/test/diagnostics.ts
+++ b/test/diagnostics.ts
@@ -5,16 +5,10 @@ import { ElmWorkspace } from "../src/elmWorkspace";
 import * as path from "path";
 import { Settings } from "../src/util/settings";
 import Parser from "web-tree-sitter";
-import {
-  getCancellationFilePath,
-  FileBasedCancellationTokenSource,
-  getCancellationFolderPath,
-  ThrottledCancellationToken,
-} from "../src/cancellation";
-import { randomBytes } from "crypto";
-import { Diagnostic } from "vscode-languageserver";
 import { spawnSync } from "child_process";
 import { appendFileSync, readFileSync } from "fs";
+import { Diagnostic } from "../src/util/types/diagnostics";
+import { performance } from "perf_hooks";
 
 container.register("Connection", {
   useValue: {
@@ -25,8 +19,8 @@ container.register("Connection", {
       warn: (): void => {
         // console.log(a);
       },
-      error: (): void => {
-        // console.log(a);
+      error: (a: string): void => {
+        console.log(a);
       },
     },
     window: {
@@ -36,6 +30,7 @@ container.register("Connection", {
     },
   },
 });
+
 container.register("Settings", {
   useValue: new Settings({} as any, {}),
 });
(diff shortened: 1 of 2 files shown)


Razzeee:
The reason to not have this on the normal build is due to it taking too long?

jmbockhorst:
Yes. I'm running it now on my personal repo and it's been going for over 2 hours. It might be froze up because I can't see the logs.
https://github.com/jmbockhorst/elm-language-server/runs/1493270378?check_suite_focus=true

We can see, but I would suspect that even if it ran smooth it would take at least 30 minutes to an hour, or possibly more. Didn't think we'd want that on every push, but we can consider.

Razzeee:
Yeah, was just wondering what too long is in this case :)

jmbockhorst:
Actually, it looks like it just keeps getting stuck. Otherwise it should complete in 11-12 minutes.

jmbockhorst:
This is ready to go now. Currently it doesn’t fail when a library fails so that we can test them. We can enable that once they are all passing or we have a known failures list.  Unfortunately it looks like there are about 150 failures, which is way more than I saw last week. I will be looking at the workspaces invididually over the next week trying to find the issues.

test/diagnostics.ts
```diff
@@ -25,8 +19,8 @@ container.register("Connection", {
       warn: (): void => {
         // console.log(a);
       },
-      error: (): void => {
-        // console.log(a);
+      error: (a: string): void => {
+        console.log(a);
```
Razzeee, line 23: Do you want to keep it like this?

test/diagnostics.ts
```diff
@@ -50,61 +45,59 @@ async function initParser(): Promise<void> {
...
     console.log(`${diagnostics.length} diagnostics found.`);
 
     diagnostics.forEach((diagnostic) => {
-      console.log(
-        `${path.basename((<any>diagnostic.data).uri)}: ${diagnostic.message}`,
-      );
+      console.log(`${path.basename(diagnostic.uri)}: ${diagnostic.message}`);
     });
 
+    console.log();
```
Razzeee, line 84: Empty log?

test/diagnostics.ts
```diff
@@ -160,26 +150,50 @@ const filteredLibs = libsToParse
     (lib) =>
       !lib.startsWith("elm/") &&
       !lib.startsWith("elm-explorations/") &&
-      !parsingFailures.includes(lib) &&
-      !compilerFailures.includes(lib) &&
+      !unknownFailures.includes(lib) &&
+      // !parsingFailures.includes(lib) &&
+      // !compilerFailures.includes(lib) &&
```
Razzeee, line 155: should these still be commented out? should we remove the variables?

Razzeee (changes requested):
Some cleanup might be needed

test/diagnostics.ts
jmbockhorst: Going to re add them once we figure out which ones are actually parser or compile errors

test/diagnostics.ts
jmbockhorst: Yeah I think it’ll help if there are any errors.

test/diagnostics.ts
jmbockhorst: Just to add a space.

PR #388, opened 2020-12-06
Typed diagnostic sources

Razzeee:


jmbockhorst (approved):
Looks good.

Razzeee:
Almost all `handlerForWorkspace` are gone. I had problems with codeLenses, that lead to the server crashing. So I reverted these. 

FileEventHandler also seemed a bit more complex.

Razzeee:
Not too sure how stable this might be, as I can't pin down the problem with codeLenses and that concerns me regarding the rest of the code.

jmbockhorst:
I think the problem with code lenses is that for the resolve request, we return the modified params we can't be JSON stringified to send to the client (because of the ElmWorkspace object).
So instead of: `const codelens = param;` we could use `const codelens: CodeLens = { range: params.range, command: params.range, data: params.data };`

src/providers/selectionRangeProvider.ts
```diff
@@ -21,20 +21,18 @@ export class SelectionRangeProvider {
...
-    elmWorkspace: IElmWorkspace,
+    params: ISelectionRangeParams,
   ): SelectionRange[] | null => {
     this.connection.console.info(`Selection Ranges were requested`);
 
     const ret: SelectionRange[] = [];
 
-    const forest = elmWorkspace.getForest();
-    const tree: Tree | undefined = forest.getTree(params.textDocument.uri);
+    const tree: Tree | undefined = params.sourceFile.tree;
```
jmbockhorst, line 35: `Tree` should always be defined here, and we can remove the subsequent check.

test/definitionProviderTests/definitionProviderTestBase.ts
```diff
@@ -99,20 +98,17 @@ export class DefinitionProviderTestBase {
...
+            position: determinedTestType.invokePosition,
+            program,
+            sourceFile,
+          });
 
-          const rootNode = this.treeParser
-            .getWorkspace(determinedTestType.sources)
-            .getForest()
-            .treeMap.get(invokeUri)!.tree.rootNode;
+          const rootNode = program.getForest().treeMap.get(invokeUri)!.tree
```
jmbockhorst, line 110: You can use `program.getSourceFile(invokeUri)` now.

jmbockhorst (approved):

PR #390, opened 2020-12-07
Fix another union constructor definition bug

jmbockhorst:
- Fixes union constructor definitions in a better way than I did before (did it how intellij-elm does)
- Show record diffs in the type error. The logic was there already, just forget to pass the diff through
- Update the diagnostics test to skip libraries with node errors and skip non-writable files (Cuts the time down to ~35 minutes)
- Also fixed a caching bug with type annotations (should fix the one you saw when making changes in `elm-cell-grid`)

We are now down to ~60 failing libraries with incorrect diagnostics.

### DIFF
--- a/src/util/types/typeExpression.ts
+++ b/src/util/types/typeExpression.ts
@@ -67,7 +67,7 @@ export class TypeExpression {
       const inferenceResult = new TypeExpression(
         e,
         workspace,
-        false,
+        /* rigidVars */ false,
       ).inferTypeDeclaration(e);
 
       TypeReplacement.freeze(inferenceResult.type);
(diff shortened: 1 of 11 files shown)


src/util/types/typeExpression.ts
```diff
@@ -67,7 +67,7 @@ export class TypeExpression {
       const inferenceResult = new TypeExpression(
         e,
         workspace,
-        false,
+        /* rigidVars */ false,
```
Razzeee, line 70: Seeing this I keep thinking we could just use an union type to get rid of the true/false. But I guess you have good reasons for this.

jmbockhorst: Union type could work fine, just doing this to increase readability.

PR #393, opened 2020-12-09
Read elm-analyse.json

jmbockhorst:


src/providers/diagnostics/elmLsDiagnostics.ts
```diff
@@ -330,32 +359,86 @@ export class ElmLsDiagnostics {
...
+    }
+
+    let elmAnalyseJson = {};
+    try {
+      elmAnalyseJson = require(path.join(
+        workspacePath,
+        "elm-analyse.json",
+      )) as IElmAnalyseJson;
+    } catch {
+      // Not found
```
Razzeee, line 435: We should probably log?

src/providers/diagnostics/elmLsDiagnostics.ts
```diff
@@ -330,32 +359,86 @@ export class ElmLsDiagnostics {
...
+          : this.getFullyAppliedOperatorAsPrefixDiagnostics(tree)),
+        ...(elmAnalyseJson.checks?.UnusedTypeAlias === false
+          ? []
+          : this.getUnusedTypeAliasDiagnostics(tree)),
+        ...(elmAnalyseJson.checks?.UnusedValueConstructor === false
+          ? []
+          : this.getUnusedValueConstructorDiagnostics(tree)),
       ];
     } catch (e) {
       console.log(e);
```
Razzeee, line 416: Can you log this via connection too?

src/providers/diagnostics/elmLsDiagnostics.ts
```diff
@@ -330,32 +359,86 @@ export class ElmLsDiagnostics {
     treeContainer: ITreeContainer,
     elmWorkspace: IElmWorkspace,
   ): IDiagnostic[] => {
+    const elmAnalyseJson = this.getElmAnalyseJson(
```
Razzeee, line 362: I'm wondering if this needs to do a findup from the folder where the file is that we're analyzing.

jmbockhorst: Maybe, but I'm not sure how often that case would occur. For performance reasons I didn't want to be checking for it every time.

Razzeee: I guess we're good, https://stil4m.github.io/elm-analyse/#/configuration seems like it's not the case.
> By default all checks are enabled. To disable checks, add an elm-analyse.json file to the root of the Elm project (beside elm.json).

PR #394, opened 2020-12-10
Update packages ranking

rlefevre:


rlefevre:
Wait before merging, I think broken packages like `Skinney` ones should be removed first.

rlefevre:
I have removed `Skinney/*` and `chelovek0v/bbase64` broken packages. I will make a script to get the full list in a future update but this should be an acceptable update already.

rlefevre:
All broken authors and packages have been removed.

Razzeee:
Thank you!

PR #396, opened 2020-12-11
Fix module resolution

jmbockhorst:
Changed a few things. The main thing is that we don't resolve modules just based on the file system (should also be faster without checking the disk). Instead we resolve accessible (exposed) modules on workspace init. There isn't a good way to a test for this right now, but I did a good amount of manual testing and the diagnostics tests will show a regression is resolving modules.

As a side note, we need a better way structure our tests. Right now, we have MockElmWorkspace, which has a lot of duplicated logic from ElmWorkspace. Ideally, we should be able to test logic in ElmWorkspace as well.


Razzeee:
Sidenote: Agreed, I'm aware of that too, I also think we might be able to share more code on the different provider tests. Especially if we start to cover some of the more simple ones like folding etc.

src/elmWorkspace.ts
```diff
@@ -99,6 +99,7 @@ export interface IElmWorkspace {
     sourceFile: ITreeContainer,
     cancellationToken?: ICancellationToken,
   ): Diagnostic[];
+  hasAccesibleModule(moduleName: string): boolean;
```
Razzeee, line 102: Typo

jmbockhorst:
This is still going to need some work. After thinking about it, each package needs a map of its accessible modules, because something may not be accessible to the root project, but a dependency might still need access.

jmbockhorst:
Finished module resolution rewrite and also got rid of MockElmWorkspace so we can test it directly. Diagnostic tests showed that some of the "applications" are using 0.19.0, but we are using 0.19.1 from npm, so they don't work correctly. Other than that, it did not show any regressions. 

Looks like there is a Windows issue with the test. Will have to look tomorrow.

Razzeee:
Seems to be caused by the drive letter being uppercased in one of the cases.

And thus this comparison not working `path.parse(uri).dir === path.parse(baseUri).dir`

Razzeee:
Using `URI.file(path.parse(uri).dir).fsPath === URI.file(path.parse(baseUri).dir).fsPath` fixes it, but I think we have to improve the design, as it's likely not just the tests but also the production code.

I would suggest not keeping `string` if possible but `URI`. For e.g. `IElmProject` has a uri, that should be of type `URI`.

jmbockhorst:
Yeah, I was thinking it would better to standardize using `URI`.

Issue #32, opened 2019-05-19
Reload imports that are outside of the project


Razzeee:
Currently we are only loading all files on import. We need to figure out a way how to do this after `elm.json` changes and `elm make` downloaded the new deps.

Razzeee:
```
  "source-directories": [
        "src",
        "../../../../Lib/elm/packages/elm-responsive-ui/src",
        "../../../../Lib/elm/packages/elm-phoenix-websocket/src"
    ],

```

this also causes problems with the files in those folders not correctly updating.

Razzeee:
We do this slightly different now, but it should cover the usecase.

PR #402, opened 2020-12-15
Watch for elm.json changes

jmbockhorst:
Not sure if the progress is working. It wasn't showing up in my test but it might have finished too fast.

### DIFF
--- a/src/elmWorkspace.ts
+++ b/src/elmWorkspace.ts
@@ -23,6 +23,7 @@ import {
   DefinitionResult,
   TypeChecker,
 } from "./util/types/typeChecker";
+import chokidar, { FSWatcher } from "chokidar";
 
 const readFile = util.promisify(fs.readFile);
 
@@ -223,6 +224,7 @@ export class ElmWorkspace implements IElmWorkspace {
   private elmPackageCache!: IElmPackageCache;
   private resolvedPackageCache = new Map<string, IElmPackage>();
   private host: IProgramHost;
+  private elmJsonWatcher: FSWatcher | undefined;
 
   constructor(private rootPath: URI, programHost?: IProgramHost) {
     this.settings = container.resolve("Settings");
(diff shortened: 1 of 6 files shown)


src/elmWorkspace.ts
```diff
@@ -385,6 +388,24 @@ export class ElmWorkspace implements IElmWorkspace {
     const pathToElmJson = path.join(this.rootPath.fsPath, "elm.json");
     this.connection.console.info(`Reading elm.json from ${pathToElmJson}`);
 
+    await this.elmJsonWatcher?.close();
+    this.elmJsonWatcher = chokidar.watch(pathToElmJson).on("change", () => {
```
Razzeee, line 392: We might want to debounce here

Razzeee:
No, I don't think #32 is closed, we still miss if the elm package cache changes.

- Checkout an elm project from git
- Open it in vscode
- The server starts
- You run `elm make` for the first time and get dependencies
- The server doesn't know about them

jmbockhorst:
Should we run `elm make` on init to get dependencies? The server won't even start otherwise.

Razzeee:
Might be a way.

I feel like the old behavior would be better, if the server can start everything and we start to watch the package folder for changes. Not a huge fan of watching a folder in such a central location, but what can we do.

src/elmWorkspace.ts
jmbockhorst: I'm not sure how much the file would be saved many times quickly. Since is just disk changes, not editor changes.

jmbockhorst:
What do you mean by old behavior?

Razzeee:
The old behavior indexed what was there (but wrongish) but didn't fail if a package was missing. It ended up without that package in it's indexes, which causes inconsistencies, but it was still useable. Describing how it was before the new package resolve algorithm.

jmbockhorst:
Alright, so maybe the dependency solver algorithm should look at the packages on Github (if online) or wherever the compiler gets them. That way it can validate the package versions like the compiler would, and get the correct version no matter what is on disk. Then we can just ignore if the package is missing on disk, like before. It would still fail if the dependency versions were messed up, just like the compiler would fail. But I think that is fine because once they fix it, we will reload the server.

jmbockhorst:
Fixed the progress reporting and tested it. This is working nicely now.

src/elmWorkspace.ts
Razzeee: Didn't know that, then again, editors like vim seem to change with every keystroke?

Razzeee:
Test suite seems to run weirdly long

src/elmWorkspace.ts
jmbockhorst: We should investigate, but I wouldn't think they would save to disk every keystroke.

Razzeee: I know that a user reported that, but it might be down to the vim config

PR #407, opened 2020-12-15
Add linked editing ranges provider

jmbockhorst:


Razzeee:
Seems to work, but we need to document it in the readme and maybe also mention the needed setting (in the vscode client)

Razzeee:
I'm actually not convinced we should add this. It will always break code like this, or we need to push the same ranges as we do for rename. Which begs the question if we shouldn't focus on one. 

It seems like the need arose from Html tags, where this makes more sense. But I can't think of such a case in elm.

jmbockhorst:
I think its just a nice to have when you are editing a function name. Even with the rename functionality, sometimes you just add a function and want to change the name manually or something. Either way, if someone does manually edit a function name, it could be broke too. I see the argument, but in my opinion, this feature doesn't hurt anything. Maybe we just shouldn't draw attention to it as a new feature to be used often.

src/providers/linkedEditingRangesProvider.ts
```diff
@@ -0,0 +1,77 @@
...
+      nodeAtPostion.type === "lower_case_identifier"
+    ) {
+      ranges.push(range);
+      ranges.push(this.addLinesToRange(range, 1));
+    }
+
+    return { ranges };
+  };
+
+  private addLinesToRange(range: Range, linesToAdd: number): Range {
```
Razzeee, line 65: In theory there could be 
```
initialModel : Model
--test
initialModel =
    { count = 0 }
```
which would break this, but I don't think that happens often enough

PR #408, opened 2020-12-16
Use package.elm-lang.org to resolve dependency versions

### LINKED ISSUE #344: Run `elm make` on server initialization to download dependencies
I think without passing an entry file path, it will just download the dependencies and quit, so it shouldn't affect startup time very much.

jmbockhorst:
First, try using package.elm-lang.org to solve dependency versions. This way, we can solve the dependencies without throwing an error if missing. It will show a notification error for any missing packages, but everything else will continue to work. Also run `elm make` on initialize, so theoretically it would never have missing packages. If we can't solve the packages (meaning they hand edited dependency versions in elm.json), the server will not start, but neither will the compiler. I think this is fine coupled with #402, because once they fix elm.json, we will try to reload. Closes #344.

### DIFF
--- a/src/elmPackageCache.ts
+++ b/src/elmPackageCache.ts
@@ -1,0 +1,113 @@
+import { readdir } from "fs";
+import * as utils from "./util/elmUtils";
+import { ElmJson } from "./elmWorkspace";
+import { promisify } from "util";
+import { IConstraint, IVersion } from "./util/elmUtils";
+
+const readDir = promisify(readdir);
+
+export interface IPackage {
+  dependencies: Map<string, IConstraint>;
+  version: IVersion;
+}
+
+export interface IElmPackageCache {
+  getVersions(packageName: string): Promise<IVersion[]>;
+  getDependencies(
+    packageName: string,
+    version: IVersion,
+  ): Promise<Map<string, IConstraint>>;
+}
+
+export class ElmPackageCache implements IElmPackageCache {
+  private versionsCache = new Map<string, IVersion[]>();
+  private dependenciesCache = new Map<string, Map<string, IConstraint>>();
+
+  constructor(
+    private packagesRoot: string,
+    private loadElmJson: (elmJsonPath: string) => Promise<ElmJson>,
+  ) {}
+
+  public async getVersions(packageName: string): Promise<IVersion[]> {
+    const cached = this.versionsCache.get(packageName);
+
+    if (cached) {
+      return cached;
+    }
+
+    const versions = await this.getVersionsFromFileSystem(packageName);
+
+    this.versionsCache.set(packageName, versions);
+
+    return versions;
+  }
+
+  public async getDependencies(
+    packageName: string,
+    version: IVersion,
+  ): Promise<Map<string, IConstraint>> {
+    const cacheKey = `${packageName}@${version.string}`;
+    const cached = this.dependenciesCache.get(cacheKey);
+
+    if (cached) {
+      return cached;
+    }
+
+    const dependencies = await this.getDependenciesFromFileSystem(
+      packageName,
+      version,
+    );
+
+    this.dependenciesCache.set(cacheKey, dependencies);
+
+    return dependencies;
+  }
+
+  private async getVersionsFromFileSystem(
+    packageName: string,
+  ): Promise<IVersion[]> {
+    const maintainer = packageName.substring(0, packageName.indexOf("/"));
+    const name = packageName.substring(
+      packageName.indexOf("/") + 1,
+      packageName.length,
+    );
+
+    const pathToPackage = `${this.packagesRoot}${maintainer}/${name}/`;
+    const folders = await readDir(pathToPackage, "utf8");
+
+    const allVersions: IVersion[] = [];
+
+    for (const folderName of folders) {
+      const version = utils.parseVersion(folderName);
+
+      if (
+        Number.isInteger(version.major) &&
+        Number.isInteger(version.minor) &&
+        Number.isInteger(version.patch)
+      ) {
+        allVersions.push(version);
+      }
+    }
+
+    return allVersions;
+  }
+
+  private async getDependenciesFromFileSystem(
+    packageName: string,
+    version: IVersion,
+  ): Promise<Map<string, IConstraint>> {
+    const elmJsonPath = `${this.packagesRoot}${packageName}/${version.string}/elm.json`;
+    const elmJson = await this.loadElmJson(elmJsonPath);
+
+    return this.parseDependencies(elmJson);
+  }
+
+  private parseDependencies(elmJson: ElmJson): Map<string, IConstraint> {
+    return new Map<string, IConstraint>(
+      Object.entries(elmJson.dependencies).map(([name, constraint]) => [
+        name,
+        utils.parseConstraint(constraint),
+      ]),
+    );
+  }
+}
--- a/src/elmWorkspace.ts
+++ b/src/elmWorkspace.ts
@@ -1,4 +1,4 @@
-import fs, { readdirSync } from "fs";
+import fs from "fs";
 import globby from "globby";
 import os from "os";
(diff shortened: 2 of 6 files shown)


test/solver.test.ts
```diff
@@ -72,14 +72,27 @@ describe("module resolution solver", () => {
...
     expected: [string, IVersion][] | undefined,
     cache = packageCache,
   ) {
-    expect(solveDependencies(cache, new Map(deps))).toEqual(
-      expected ? new Map(expected) : undefined,
-    );
+    expect(
+      await solveDependencies(
+        {
+          getVersions: (pacakgeName) =>
```
Razzeee, line 83: Typo

src/elmWorkspace.ts
```diff
@@ -391,13 +315,34 @@ export class ElmWorkspace implements IElmWorkspace {
...
         this.loadElmJson.bind(this),
       );
+      this.missingPackages = new Set();
       this.rootProject = await this.loadRootProject(pathToElmJson);
       this.forest = new Forest(this.rootProject);
 
+      this.missingPackages.forEach((missingPackage) => {
+        this.connection.window.showErrorMessage(
+          `Could not load Elm package ${missingPackage}. 
+         Trying running \`elm make\` to install dependencies.`,
```
Razzeee, line 342: Not a native speaker, but shouldn't this be `Try running ...` as we want the user to do it manually?

jmbockhorst: Yep, another typo 😬

src/elmPackageCache.ts
```diff
@@ -0,0 +1,152 @@
...
+  constructor(
+    private packagesRoot: string,
+    private loadElmJson: (elmJsonPath: string) => Promise<ElmJson>,
+  ) {
+    this.packagesFromWebsite = this.getJsonFromUrl<{
+      [packageName: string]: string[];
+    }>(`${this.packageUrl}/all-packages`);
+  }
+
+  public async getVersions(packageName: string): Promise<IVersion[]> {
```
Razzeee, line 39: I think we should be getting the dep constraint here and try to check on disk first (be nice to the servers)

The other thing is, that we might end up just asking the website then only return these versions and our constraint might not even match. So I think we need to check it here, see if we can solve with local data and if not, ask the website.

src/elmPackageCache.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    if (cached) {
+      return cached;
+    }
+
+    const dependencies =
+      (await this.getDependenciesFromFileSystem(packageName, version)) ??
+      (await this.getDependeciesFromPackageWebsite(packageName, version));
+
+    this.dependenciesCache.set(packageName, dependencies);
```
Razzeee, line 69: It might be a problem, that our key is only the packageName and when it should be package name and version?

src/elmPackageCache.ts
```diff
@@ -0,0 +1,152 @@
...
+  ): Promise<IVersion[] | undefined> {
+    try {
+      const packages = await this.packagesFromWebsite;
+      return packages[packageName].map(utils.parseVersion);
+    } catch (e) {
+      // No internet connection
+    }
+  }
+
+  private async getDependeciesFromPackageWebsite(
```
Razzeee, line 85: Typo

src/elmPackageCache.ts
```diff
@@ -0,0 +1,152 @@
...
+    packageName: string,
+    version: IVersion,
+  ): Promise<Map<string, IConstraint>> {
+    console.log(`Loading deps of ${packageName}@${version.string}`);
```
Razzeee, line 89: Log to connection?

src/elmPackageCache.ts
```diff
@@ -0,0 +1,152 @@
...
+    } catch {
+      // File doesn't exist
+    }
+  }
+
+  private parseDependencies(elmJson: ElmJson): Map<string, IConstraint> {
+    return new Map<string, IConstraint>(
+      Object.entries(elmJson.dependencies).map(([name, constraint]) => [
+        name,
+        utils.parseContraint(constraint),
```
Razzeee, line 144: Typo

src/elmPackageCache.ts
jmbockhorst: This is just to get a list of all the versions for a package. We only ask the website for this once per workspace init. Checking disk first wouldn't help because it might not be complete.

Razzeee:
Am I getting this right, while we now are able to resolve the dep tree, we might end up not having the deps on disk and thus not indexing them/offering functionality based on it?

src/elmPackageCache.ts
Razzeee: Yeah, but in that case we would ask the server. So looking at the disk first doesn't hurt.

jmbockhorst:
Correct, but since we are running elm make beforehand, every package should be there. And I don't think we want to resolve a different version (whatever they have on disk) of a package than the compiler would, giving different type inference, etc.

src/elmPackageCache.ts
jmbockhorst: How will we know if we are missing a version on disk if we don't check the website?

src/elmPackageCache.ts
jmbockhorst: Meant to remove that.

src/elmPackageCache.ts
jmbockhorst: Good catch.

src/elmPackageCache.ts
Razzeee: You mean for libs probably due to the ranges? If we have something that satisfies the range, it should be fine to us. Especially, if we have run elm make otherwise we might even use newer versions then elm make is using?

Razzeee:
We should test this without a network connection.

src/elmPackageCache.ts
jmbockhorst: Well I think the compiler always grabs the latest version even if a version already satisfies a constraint. This is the behavior I was trying to follow. I will to do additional testing to verify the behavior of the compiler.

jmbockhorst:
I did test that and it works as long as a version on disk satisfies the constraints.

src/elmPackageCache.ts
Razzeee: I would expect the compiler to do some kind of caching. Maybe via that since endpoint? https://package.elm-lang.org/all-packages/since/8409
And I'm not sure if we can manage to always agree with that. And I think we hit the servers twice now, once via elm make and at least once to get all-packages via axios.

jmbockhorst: For clarity, I just checked, and here is what the compiler does in pseudocode:
```
if hasRegistryCache then
   update the registry, using /all-packages/{size}, where size is the last known size of the registry. This route gets all packages added since that size
         if success then
              update the old registry with the added packages
         else
              if no internet, return the cache

else
    fetch from /all-packages
         if success then
              return the registry
         else
              fail if no internet
```

That being said, since we are running `elm make`, I guess we have no need of checking the website at all. We just assume everything is there as it should be. 

We need to think about all the scenarios:
- **Elm not installed, elm make doesn't run**: This is the one case where looking at the website would do us good, but how often will it occur? If we don't look at the website, we might not be able to solve it (versions are missing on disk), and the the server won't start. But then again, if we did look at the website, the server could start, but probably missing all dependencies anyways, so wouldn't do much good.
- **If there is no internet**, we will look at disk no matter what. If we can't solve, then we should fail anyways. This seems like a rare case as well. 
- **elm make runs and gets latest/correct dependencies**: This seems like the most common scenario, and we wouldn't need to look on the website. 

So, the only change we need to make here is running `elm make` on init, and that should solve any problems with not being able to solve. Anything seems like a rare case, so I am fine to remove all logic that checks the website.

Razzeee: Agreed

PR #409, opened 2020-12-16
Adopt file events APIs

jmbockhorst:
Adopt the new file events APIs and adjusted some behavior for module renaming so that it works consistently now. `Module.App.func` references are now included as a module reference for `Module.App`, thus being renamed if the module/file is renamed. Also, if you try to rename on the `Module` or `App` part, you will be renaming the module `Module.App` (different from before). Renaming the function part will still rename the function. 

Also added the ability to rename a directory and have all containing files updated.

### DIFF
--- a/src/elmWorkspace.ts
+++ b/src/elmWorkspace.ts
@@ -78,8 +78,8 @@ interface IElmPackageJson {
 export interface IElmWorkspace {
   init(progressCallback: (percent: number) => void): void;
   hasDocument(uri: URI): boolean;
-  hasPath(uri: URI): boolean;
-  getPath(uri: URI): string | undefined;
+  isInSourceDirectory(uri: string): boolean;
+  getSourceDirectoryOfFile(uri: string): string | undefined;
   getSourceFile(uri: string): ITreeContainer | undefined;
   getForest(synchronize?: boolean): IForest;
   getRootPath(): URI;
@@ -175,15 +175,16 @@ export class ElmWorkspace implements IElmWorkspace {
     return !!this.forest.getTree(uri.toString());
   }
 
-  public hasPath(uri: URI): boolean {
-    return !!this.getPath(uri);
+  public isInSourceDirectory(uri: string): boolean {
+    return !!this.getSourceDirectoryOfFile(uri);
   }
 
-  public getPath(uri: URI): string | undefined {
+  public getSourceDirectoryOfFile(uri: string): string | undefined {
+    uri = normalizeUri(uri);
     return [
       ...this.rootProject.sourceDirectories,
       ...this.rootProject.testDirectories,
-    ].find((elmFolder) => uri.fsPath.startsWith(elmFolder));
+    ].find((elmFolder) => uri.includes(elmFolder));
   }
 
   public getSourceFile(uri: string): ITreeContainer | undefined {
@@ -588,11 +589,7 @@ export class ElmWorkspace implements IElmWorkspace {
     (await this.host.readDirectory(sourceDir)).forEach((matchingPath) => {
       matchingPath = normalizeUri(matchingPath);
 
-      const moduleName = path
-        .relative(sourceDir, matchingPath)
-        .replace(".elm", "")
-        .split("/")
-        .join(".");
+      const moduleName = utils.getModuleName(matchingPath, sourceDir);
 
       project.moduleToUriMap.set(moduleName, URI.file(matchingPath).toString());
 
--- a/src/elmWorkspace.ts
+++ b/src/elmWorkspace.ts
@@ -184,7 +184,7 @@ export class ElmWorkspace implements IElmWorkspace {
     return [
       ...this.rootProject.sourceDirectories,
       ...this.rootProject.testDirectories,
-    ].find((elmFolder) => uri.includes(elmFolder));
+    ].find((elmFolder) => uri.startsWith(elmFolder));
   }
 
   public getSourceFile(uri: string): ITreeContainer | undefined {
--- a/src/providers/handlers/fileEventsHandler.ts
+++ b/src/providers/handlers/fileEventsHandler.ts
@@ -158,7 +158,9 @@ export class FileEventsHandler {
     file: string,
     elmWorkspace: IElmWorkspace,
   ): string | undefined {
-    const sourceDir = elmWorkspace.getSourceDirectoryOfFile(file);
+    const sourceDir = elmWorkspace.getSourceDirectoryOfFile(
+      URI.parse(file).fsPath,
+    );
 
     // The file is not in a source dir (shouldn't happen)
     if (!sourceDir) {
--- a/src/providers/handlers/fileEventsHandler.ts
+++ b/src/providers/handlers/fileEventsHandler.ts
@@ -1,112 +1,199 @@
-import { relative } from "path";
-import { IElmWorkspace } from "../../elmWorkspace";
+import { container } from "tsyringe";
 import {
-  OnDidCreateFilesRequest,
-  OnDidRenameFilesRequest,
-} from "../../protocol";
+  Connection,
+  CreateFilesParams,
+  DeleteFilesParams,
+  FileCreate,
+  FileDelete,
+  FileRename,
+  RenameFilesParams,
+  WorkspaceEdit,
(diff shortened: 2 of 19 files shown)


src/elmWorkspace.ts
```diff
@@ -175,15 +175,16 @@ export class ElmWorkspace implements IElmWorkspace {
...
   }
 
-  public getPath(uri: URI): string | undefined {
+  public getSourceDirectoryOfFile(uri: string): string | undefined {
+    uri = normalizeUri(uri);
     return [
       ...this.rootProject.sourceDirectories,
       ...this.rootProject.testDirectories,
-    ].find((elmFolder) => uri.fsPath.startsWith(elmFolder));
+    ].find((elmFolder) => uri.includes(elmFolder));
```
Razzeee, line 187: Why includes? Shouldn't startsWith be better?

jmbockhorst: At some point there was a case that needed this, but I don't remember. I can check again to see if it's still needed.

src/util/types/typeChecker.ts
```diff
@@ -29,6 +29,7 @@ import { Utils } from "../utils";
 import { TypeExpression } from "./typeExpression";
 import { ICancellationToken } from "../../cancellation";
 import { Diagnostic, Diagnostics, error } from "./diagnostics";
+import { node } from "execa";
```
Razzeee, line 32: Seems unused

test/definitionProviderTests/moduleResolveDefinition.test.ts
```diff
@@ -133,7 +133,7 @@ import Module
 import Module.Model
       
 func = Module.Model.func
-      --^Module.elm
+      --^Module/Model.elm
```
Razzeee, line 136: This is confusing to me, can you double check if it's correct, I will try to do the same.

src/providers/renameProvider.ts
```diff
@@ -118,19 +132,37 @@ export class RenameProvider {
     newName: string,
   ): [{ [uri: string]: TextEdit[] }, TextDocumentEdit[]] {
     const edits: { [uri: string]: TextEdit[] } = {};
+    let originalName = affectedNodes?.originalNode.text ?? "";
+
+    // Adjust the case of Module.App.func
```
Razzeee, line 137: I think this comment could be better? `Helps us to rename fully qualified functions without changing the last part`

test/definitionProviderTests/moduleResolveDefinition.test.ts
jmbockhorst: I changed the behavior to be like this. Since it doesn't make any sense in this context to navigate to a module named `Module`, which may or may not exist (if it doesn't, it navigates to `func`, which also doesn't make sense), it now takes the entire module name `Module.Model` when getting the definition. This was to improve renames, so it will consistently rename the the entire module `Module.Model` if the cursor is in either part.

Razzeee: But why is the `--X` unmoved? Shouldn't the test fail?

jmbockhorst: Hmm your right. I think it’s because the —X just determines a Position, not the source file. In this case the position is the same still.

src/providers/handlers/fileEventsHandler.ts
```diff
@@ -1,112 +1,201 @@
...
-    newFile: URI,
-    elmWorkspace: IElmWorkspace,
-  ): Promise<void> {
-    if (!newFile.toString().endsWith(".elm")) {
-      return;
+  private onWillRenameFile({
+    oldUri,
+    newUri,
+    program,
+    sourceFile,
```
Razzeee, line 101: I think we should mark sourceFile as optional in IParams

src/providers/handlers/fileEventsHandler.ts
```diff
@@ -1,112 +1,201 @@
...
   }
 
-  getModuleNameFromFile(
-    file: URI,
+  private onWillDeleteFile({ uri, program }: IDeleteFileParams): void {
+    program.getForest().removeTree(uri);
+  }
+
+  private getModuleNameFromFile(
+    file: string,
```
Razzeee, line 158: I think we would normally have called this uri?

test/renameProvider.test.ts
```diff
@@ -0,0 +1,201 @@
...
+func = Module.App.foo
+                 --^
+
+--@ Module/App.elm
+module Module.App exposing (foo)
+
+foo = ""		
+`;
+
+    const renameRange = Range.create(
```
Razzeee, line 120: Seems like intelliJ-elm does this via {-carret-} which is nice, but not sure why they do it like that. The reference implementation in the rust world also does the same.

Can't think of good reasons, it might be a bit clearer where the cursor is, as we end up always pointing at a char, but the cursor in the real case will be between chars.
A good reason for us might be to reuse tests from intelliJ, but it's only 12 tests anyway.

https://github.com/klazuka/intellij-elm/blob/master/src/test/kotlin/org/elm/ide/refactoring/ElmRenameTest.kt
https://github.com/intellij-rust/intellij-rust/blob/master/src/test/kotlin/org/rust/ide/refactoring/RenameTest.kt

test/renameProvider.test.ts
```diff
@@ -0,0 +1,201 @@
...
+      Position.create(4, 18),
+      Position.create(4, 21),
+    );
+    await testPrepareRename(source, renameRange);
+
+    const newName = "bar";
+    await testRename(source, newName, [
```
Razzeee, line 127: It would be much nicer to not have to write a text edit and just compare the strings that end up in each file?

See the implementation in the intelliJ repo. They basically hand over a string for a file with a {-caret-} and a string how it's expected to look after renaming.

Razzeee: I guess that might be a problem, whenever we want to test if renames in multiple files were successful.

src/elmWorkspace.ts
Razzeee: Please do :)

Razzeee (changes requested):
Some minors in the implementation, but I think we can still improve the tests.

test/definitionProviderTests/moduleResolveDefinition.test.ts
Razzeee: Yeah, it's right by mistake :/

test/renameProvider.test.ts
jmbockhorst: I was thinking this too, just didn't get to it. But I guess we need some way to apply text edits to a the string sources. Unless something exists that is easier, we'll probably have to convert our sources to line/col buffers and grab what vscode does to apply edits.

Razzeee: Your right, I was hoping that we somehow could use `this.connection.workspace.applyEdit` but seems unlikely. I guess it's fine to create an issue from this and move on.

src/providers/handlers/fileEventsHandler.ts
Razzeee: Seems to be a bigger refactor, probably something for later

jmbockhorst: Yeah, I’d rather not, because it will almost only exist.

test/renameProvider.test.ts
jmbockhorst: Well this range is the expected range of the rename, not just the caret position.

test/renameProvider.test.ts
jmbockhorst: Was able to add a simple implementation that applies code edits. Certainly room for improvement in the future but it works for now.

src/elmWorkspace.ts
jmbockhorst: I changed it back and it seems to be fine. Only tested manually on MacOS though. Since the tests are passing on Windows, I'm fairly confident it should be fine.

jmbockhorst:
With the new `applyEditsToSource`, in the future we should be able to easily add tests for code actions fairly easily.

Razzeee:
Thanks 🎉

PR #411, opened 2020-12-18
Run elm make on init and clean up dependency solver

### LINKED ISSUE #344: Run `elm make` on server initialization to download dependencies

jmbockhorst:
Better alternative to #408. Closes #344.


Razzeee:
Seems like the diagnostic test needs an update

jmbockhorst:
For some reason, the full diagnostics tests ran when they shouldn't have.

PR #418, opened 2020-12-22
Start work on 2.0.0 changelog

jmbockhorst:
Since there are a lot of changes, I tried to break them up in sections and add more description.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,34 @@
 # Change Log
 
+## 2.0.0
+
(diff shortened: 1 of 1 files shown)


Razzeee:
I pushed an improved proposal

Razzeee:
I think we're missing the file api changes in our readme? https://github.com/elm-tooling/elm-language-server#features (not just under features)

jmbockhorst:
Yeah, I had never added them.

Issue #282, opened 2020-05-28
Exposing/unexposing for ports


jmbockhorst:
Ports should have expose/unexpose code actions, code lenses, and completions where necessary, just like functions and types.

PR #430, opened 2020-12-30
Handle goto/find ref and rename of ports

### LINKED ISSUE #355: Can not refactor/rename port fn?
## Expected Behavior
Able to rename port fn

## Current Behavior
image

## Your Environment
* Version used: latest
* Editor used: Vscode
* Environment name and version (e.g. node.js 5.4):
* Operating System and version:

Razzeee:
Closes #355

### DIFF
--- a/src/providers/codeLensProvider.ts
+++ b/src/providers/codeLensProvider.ts
@@ -268,38 +268,14 @@ export class CodeLensProvider {
         node.type === "type_declaration" ||
         node.type === "type_alias_declaration"
       ) {
-        const typeNode = TreeUtils.findFirstNamedChildOfType(
-          "upper_case_identifier",
-          node,
-        );
+        const typeNode = node.childForFieldName("name");
 
         if (typeNode) {
           codeLens.push(this.createReferenceCodeLens(typeNode, uri));
         }
       }
-    });
 
-    TreeUtils.descendantsOfType(tree.rootNode, "value_declaration").forEach(
-      (node) => {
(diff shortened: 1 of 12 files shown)


jmbockhorst (approved):
Works good. Probably should add a few tests though. 

I found a few consistency issues, but they don't necessarily have to be this PR:
- We should probably have codelenses for ports like functions
- Hovering over the port name should show the type information, like functions

Razzeee:
Some of those are tracked in https://github.com/elm-tooling/elm-language-server/issues/282

Will look at the codelenses part and how easy hover would be.

Razzeee:
Added codelenses

src/providers/codeLensProvider.ts
```diff
@@ -289,6 +298,18 @@ export class CodeLensProvider {
       },
     );
 
+    TreeUtils.descendantsOfType(tree.rootNode, "port_annotation").forEach(
```
jmbockhorst, line 301: `descendantsOfType` is pretty bad for performance, so best to avoid it when we can.

jmbockhorst: Just wanted to clarify, calling `descendantsOfType` on the root node is what can be bad. Using it in contexts like `import_clause` would not be a problem, since you know there are not many descendants to traverse.

jmbockhorst (approved):

src/providers/codeLensProvider.ts
Razzeee: I think we should just not use it, if we know, that we only want to search in the immediate children (as it was here the case, but I copied without thinking)

PR #432, opened 2020-12-31
Show port type annotation in hovers

jmbockhorst:
It won't let me push to your branch, so I had to create a PR into that branch.


Razzeee:
I'm not sure if getting rid of `functionName` will break something - played with that too, but thought it wasn't worth the risk for this release. But let's just do it.

PR #433, opened 2020-12-31
Only show imports from direct dependencies and separate test dependencies

### LINKED ISSUE #49: Separate tests from normal imports/forest
At the moment features like completions etc will make it seem like they work if you import the correct things, even if that is from the testing dependencies or vice versa.

jmbockhorst:
Fixes #49.

### DIFF
--- a/src/elmWorkspace.ts
+++ b/src/elmWorkspace.ts
@@ -33,6 +33,7 @@ interface IElmFile {
   path: string;
   maintainerAndPackageName?: string;
   project: ElmProject;
+  isTestFile: boolean;
 }
 
 export type ElmJson = IElmApplicationJson | IElmPackageJson;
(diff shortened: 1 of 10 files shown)


jmbockhorst:
Currently `testModuleToUriMap` contains everything in `moduleToUriMap`. I did this for simplicity, but this seems like a big waste of memory. I will be looking into a better solution.

jmbockhorst:
This should be ready now.

Razzeee (changes requested):
Looks good, would like some tests, but I guess that might be hard, due to our tests not having an elm.json

Razzeee (approved):

jmbockhorst:
I was thinking about that. In sourceTreeParser we define the elm.json for tests. And we could define sources in that test directory. I should be able to get something basic.

Razzeee:
> I was thinking about that. In sourceTreeParser we define the elm.json for tests. And we could define sources in that test directory. I should be able to get something basic.

Yeah, something basic would be good, don't think there is much value in testing more then if we're loading the right deps (and not too much)

PR #437, opened 2021-01-02
Add install package code action when import is missing

### LINKED ISSUE #434: Offer to install package if you try to import a module from it
I had the idea of offering code actions to install packages when there is a missing import (module it not in direct dependencies). To do this, we would have to search through all available packages and find all packages that expose a module by that name. Suggestions would be sorted by the package ranking if there are multiple. 

The question is how can we achieve this in a performant way and without having to hit the servers much. I don't think there is a good way right now, we would need a new api from the packages website. 

I think it might still be beneficial to search your the local packages cache instead, because most likely if you are trying to import a module that isn't installed, you have used it before.

jmbockhorst:
Fixes #434. The current strategy is load all local packages from the cache on workspace init, after a 5 second delay. This happened pretty fast for me, even with a large cache. Even if it did take more time, it shouldn't affect performance very much since it is all async. I initially had a progress indicator show, but it was too quick to be worth it. 

I also have it pass a `y\n` input to the command so that is automatically says yes to the cmd prompt. We may want to think about this more. I think we should definitely show a notification of what the command did. Maybe we should even run the command first without the `y\n` to see what it will do, then prompt the user with a notification with a yes/no choice.

### DIFF
--- a/src/elmPackageCache.ts
+++ b/src/elmPackageCache.ts
@@ -3,6 +3,8 @@ import * as utils from "./util/elmUtils";
 import { ElmJson } from "./elmWorkspace";
 import { promisify } from "util";
 import { IConstraint, IVersion } from "./util/elmUtils";
+import { MultiMap } from "./util/multiMap";
+import * as path from "./util/path";
 
 const readDir = promisify(readdir);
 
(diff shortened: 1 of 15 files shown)


src/util/elmUtils.ts
```diff
@@ -1,4 +1,4 @@
-import execa, { ExecaSyncReturnValue } from "execa";
+import execa, { ExecaReturnValue, ExecaSyncReturnValue } from "execa";
```
Razzeee, line 1: Seems like a wrong merge

src/providers/codeAction/installPackageCodeAction.ts
```diff
@@ -0,0 +1,75 @@
...
+    const program = new ElmWorkspaceMatcher((uri: string) =>
+      URI.parse(uri),
+    ).getProgramFor(uri);
+
+    const clientSettings = await settings.getClientSettings();
+
+    try {
+      const result = execCmdSync(
+        clientSettings.elmPath,
+        "elm",
```
Razzeee, line 64: We might want to use elm-json for this, as I think @jfmengels is doing for elm-review and there might be benefits?

jmbockhorst: We'll have to see. I did think about that, but then saw this:

> dependencies-cache/: Caching of the dependencies of the project's configuration computed by elm-json. elm-json is a bit slow, and doesn't work great offline.

From here: https://github.com/jfmengels/node-elm-review/blob/master/documentation/tooling-integration.md

jmbockhorst:
This is pretty good now. It shows information prompts just like the command line and asks for yes/no. I think I might add a setting that allows running the install without the prompt, since that may be desirable sometimes.

jmbockhorst:
I change the cache in `ElmPackageCache` to be static, so it will live beyond each server restart. This should save some time for server restarts.

src/util/settings.ts
```diff
@@ -24,6 +25,7 @@ export class Settings {
     elmTestPath: "",
     trace: { server: "off" },
     disableElmLSDiagnostics: false,
+    skipInstallPackageConfirmation: false,
```
Razzeee, line 28: Should be added to the readme

src/elmPackageCache.ts
```diff
@@ -17,27 +19,46 @@ export interface IElmPackageCache {
...
-  private versionsCache = new Map<string, IVersion[]>();
-  private dependenciesCache = new Map<string, Map<string, IConstraint>>();
+  private static versionsCache = new Map<string, IVersion[]>();
+  private static dependenciesCache = new Map<
+    string,
+    Map<string, IConstraint>
+  >();
+  private static moduleToPackages = new MultiMap<string, string>();
+
+  private static _packagesRoot: string;
```
Razzeee, line 33: Why is this prefixed with `_` I know that's either used for private fields or not used fields in some languages.

jmbockhorst: Private field in this case.

src/util/settings.ts
jmbockhorst: Done.

src/elmPackageCache.ts
Razzeee: Then the other fields should use the same convention?

jmbockhorst: I guess. It's not a convention that I love, just needed to have a custom setter here.

Razzeee:
Can you add this to the changelog?

PR #439, opened 2021-01-03
Update deps

Razzeee:
Just some eslint cleanup


src/util/types/typeChecker.ts
```diff
@@ -977,7 +976,7 @@ export function createTypeChecker(workspace: IElmWorkspace): TypeChecker {
     ).diagnostics.forEach((diagnostic) => diagnostics.add(diagnostic));
   }
 
-  function getSourceFileOfNode(node: SyntaxNode): ITreeContainer {
-    return forest.getByUri(node.tree.uri)!;
+  function getSourceFileOfNode(node: SyntaxNode): ITreeContainer | undefined {
```
jmbockhorst, line 979: If you prefer this, that's fine. But it should always be defined, and it adds unnecessary checking since it returns undefined. 

Maybe we could just throw an error if not defined, and then we could just return `ITreeContainer`?

Razzeee:
Would be fine with that, just want to get rid of unnecessary !

src/util/types/typeChecker.ts
Razzeee: Changed

jmbockhorst (approved):

PR #442, opened 2021-01-10
Use elm-tooling-cli to install third party bins

Razzeee:
This already seems pretty good to me, probably still some corner cases left.

- [ ] we can't install a test runner like this, yet - will have to wait for `elm-test-rs`
- [x] can we just call install every time? should we check if we can already find the bin and skip or check if the versions matches?
- [x] update readme
- [x] error when download fails

### DIFF
--- a/src/elmWorkspace.ts
+++ b/src/elmWorkspace.ts
@@ -369,6 +369,7 @@ export class ElmWorkspace implements IElmWorkspace {
         try {
           await runElmToolingCliInstall(options);
         } catch (error) {
+          this.connection.window.showErrorMessage(`${error}`);
           this.connection.console.error(`Unexpected error ${error}`);
         }
       } else {
@@ -384,7 +385,7 @@ export class ElmWorkspace implements IElmWorkspace {
               const initExitCode = await elmToolingCli(["init"], options);
 
               if (initExitCode !== 0) {
-                throw "elm-tooling  init failed";
+                throw "Creating a new 'elm-tooling.json' failed";
               }
 
               await runElmToolingCliInstall(options);
@@ -411,6 +412,7 @@ export class ElmWorkspace implements IElmWorkspace {
             }
           })
           .catch((error) => {
+            this.connection.window.showErrorMessage(`${error}`);
             this.connection.console.error(`Unexpected error "${error}"`);
           });
       }
@@ -812,6 +814,6 @@ async function runElmToolingCliInstall(options: {
 }): Promise<void> {
   const installExitCode = await elmToolingCli(["install"], options);
   if (installExitCode !== 0) {
-    throw "elm-tooling install failed";
+    throw "Failed to install one of the dependencies";
   }
 }
--- a/src/elmWorkspace.ts
+++ b/src/elmWorkspace.ts
@@ -27,6 +27,7 @@ import {
 } from "./util/types/typeChecker";
 import chokidar from "chokidar";
 import { CommandManager } from "./commandManager";
+import elmToolingCli from "elm-tooling";
 
 const readFile = util.promisify(fs.readFile);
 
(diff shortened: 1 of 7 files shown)


Razzeee:
This should basically allow us to get rid of the prerequisites in https://github.com/elm-tooling/elm-language-server#requirements so that new users only need to install the extension

src/elmWorkspace.ts
```diff
@@ -363,6 +352,85 @@ export class ElmWorkspace implements IElmWorkspace {
...
+        "elm-tooling.json",
+      );
+      const options = {
+        cwd: this.rootPath.fsPath,
+        env: process.env,
+        stdin: process.stdin,
+        stderr: process.stderr,
+        stdout: process.stdout,
+      };
+      if (fs.existsSync(pathToElmToolingJson)) {
```
jmbockhorst, line 367: We should add a `fileExists` method to `IProgramHost` to use here and move this down to `createProgramHost`.

src/elmWorkspace.ts
```diff
@@ -363,6 +352,85 @@ export class ElmWorkspace implements IElmWorkspace {
...
+          await runElmToolingCliInstall(options);
+        } catch (error) {
+          this.connection.window.showErrorMessage(`${error}`);
+          this.connection.console.error(`Unexpected error ${error}`);
+        }
+      } else {
+        this.connection.console.info(`elm-tooling.json does not exist`);
+        this.connection.window
+          .showInformationMessage(
+            "Create elm-tooling.json, to store elm tool versions and download them?",
```
jmbockhorst, line 379: I'm not sure about this. We are basically forcing users to create an elm-tooling.json, even if they already have their environment setup how they like and working.

src/elmWorkspace.ts
```diff
@@ -363,6 +352,85 @@ export class ElmWorkspace implements IElmWorkspace {
...
+              await this.connection.window.showDocument({
+                uri: pathToElmToolingJson,
+              });
+
+              // we might want to restart, as the used packages might not match the elm version
+              // or we had no elm before and probably were crashing
+              await this.connection.window
+                .createWorkDoneProgress()
+                .then((progress) => {
+                  progress.begin("Restarting Elm Language Server", 0);
```
jmbockhorst, line 402: Should make a `restart` method that so this is duplicated from above.

jmbockhorst (commented):
Maybe we could only offer to create a elm-tooling.json if we can't get the elm version?

 I don't know, I'm split on this. On one hand, it would add visibility to the tool and more people would end up using it, and it's just a notification (I get plenty of bothersome notifications in vscode). On the other hand, we don't want users to feel forced to use it. 

If we do always ask, saying "No" to create elm-tooling.json should set `useElmToolingJsonForTools: true`, so we don't ask again.

src/elmWorkspace.ts
Razzeee: I did try, but the above was using it via then, which didn't work for me here, as I need it to block I guess? (might have been the `void`)

src/elmWorkspace.ts
Razzeee: Yeah, it's very opinionated, but that's basically what I'm after. But all I wanted to achive is to cut down on setup, which I'm not even sure this ends up doing in the end. I think relying on node now is the real problem for me here.

Razzeee:
> Maybe we could only offer to create a elm-tooling.json if we can't get the elm version?
> 
> I don't know, I'm split on this. On one hand, it would add visibility to the tool and more people would end up using it, and it's just a notification (I get plenty of bothersome notifications in vscode). On the other hand, we don't want users to feel forced to use it.
> 
> If we do always ask, saying "No" to create elm-tooling.json should set `useElmToolingJsonForTools: true`, so we don't ask again.

I guess that would be sensible and maybe have a flag to force it to use the cli.

I couldn't find a way to write to the settings in the language server api, but maybe I overlooked it.

PR #447, opened 2021-01-13
Compiler api refactoring

jmbockhorst:
Renames:
- `IElmWorkspace` -> `IProgram`
- `ITreeContainer` -> `ISourceFile`

Moves all compiler related code into `./src/compiler` that can be exported as a npm package. 

Here's a few things that will need to happen before we can ship this api:
- Remove all LSP specific stuff from these files. (`Program` uses `connection` a lot)
- Refine the api: What should be exposed, what should be internal, what should the language server do instead 
- Discuss tradeoffs of moving the compiler api to a separate package. (Can't use `/** @internal */` methods, but wouldn't have dependency on LSP)

~~The name "analyzer" is still up for discussion.~~ I have decided that using "compiler" is a better name. The API will probably be shipped through the `elm-language-server` package, so we can use "compiler" internally.


Razzeee:
Code looks good, but I haven't reviewed all. It needs a rebase, and I'm not sold on the compiler naming, but if we want to change that it should be a follow up.

Razzeee (approved):
Lgtm

jmbockhorst:
Yeah, we can always rename as we structure the api better.

PR #487, opened 2021-01-25
Improve unused import diagnostics

### LINKED ISSUE #486: Unused import diagnostics improvements
If there is an unused import alias and there are no exposing values, we can say the entire import is unused. Also make sure the code actions work for all cases.

jmbockhorst:
Cleans up unused import diagnostics and adds a code action to remove unused import. Closes #486. 

With this new way, if `import Html as H` is unused, it is just an unused import, and the whole thing can be removed. If the alias in `import Html as H exposing (..)` is unused, we report an unused alias, because the import is still used.

### DIFF
--- a/test/diagnosticTests/elmLsDiagnostics.test.ts
+++ b/test/diagnosticTests/elmLsDiagnostics.test.ts
@@ -459,6 +459,7 @@ foo = (+) 1 2
       ]);
     });
 
+    // This case is handled by unused_import
     it("unused alias but no exposing", async () => {
       const source = `
 module Foo exposing (..)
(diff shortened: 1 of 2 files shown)


test/diagnosticTests/elmLsDiagnostics.test.ts
```diff
@@ -451,6 +459,18 @@ foo = (+) 1 2
...
+    it("unused alias but no exposing", async () => {
+      const source = `
+module Foo exposing (..)
+
+import Bar as B
+
+foo = 1
+			`;
+
+      await testDiagnostics(source, "unused_alias", []);
```
Razzeee, line 471: I think a comment pointing out, that this will instead be catched by the logic for the whole import being unused would be helpful. (not just here)

PR #488, opened 2021-01-25
Don't crash server if it fails to initialize

### LINKED ISSUE #485: Do better error catching on initialization
Currently we crash the entire server if just one workspace fails to init. We should be able to show an error message (in elm.json for example) and still load the other workspaces.

jmbockhorst:
I found the source of the crash. Now individual requests will fail, but the entire server will not crash. Closes #485.

I was trying to send error diagnostics to `elm.json`, but it was not working. Do you think because it is not an `.elm` file, VSCode doesn't accept them?

### DIFF
--- a/src/providers/diagnostics/diagnosticsProvider.ts
+++ b/src/providers/diagnostics/diagnosticsProvider.ts
@@ -146,6 +146,10 @@ export class DiagnosticsProvider {
         );
       } else {
         this.workspaces.forEach((workspace) => {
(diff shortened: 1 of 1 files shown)


Razzeee:
I'm pretty sure, that used to work. Don't we fallback for some other errors to elm.json? I remember there being some compiler errors like that.

jmbockhorst:
Alright, well something right look into in the future.

PR #491, opened 2021-01-26
Separate elm-tooling.json handling

Razzeee:
This separates the elm-tooling.json handling, preparing us to load more values.
Should also allow starting to cache these calls at some point.

### DIFF
--- a/src/elmToolingJsonManager.ts
+++ b/src/elmToolingJsonManager.ts
@@ -1,0 +1,81 @@
+import path from "path";
+import { URI } from "vscode-uri";
+import { ITreeContainer } from "./forest";
+import { NonEmptyArray } from "./util/utils";
+import util from "util";
+import * as fs from "fs";
+
+const readFile = util.promisify(fs.readFile);
+
+export class ElmToolingJsonManager {
+  public async getEntrypoints(
+    workspaceRootPath: string,
+    sourceFile: ITreeContainer,
+  ): Promise<[NonEmptyArray<string>, string]> {
+    const elmToolingPath = path.join(workspaceRootPath, "elm-tooling.json");
+    const defaultRelativePathToFile = path.relative(
+      workspaceRootPath,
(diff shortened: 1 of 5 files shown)


src/elmToolingJsonManager.ts
```diff
@@ -0,0 +1,81 @@
...
+        }
+      } else {
+        throw new Error(`There is no "entrypoints" field.`);
+      }
+    } else {
+      throw new Error(
+        `Expected a JSON object but got: ${JSON.stringify(json)}`,
+      );
+    }
+  }
```
lydell, line 80: I haven’t tested this, but if we want to clean the code up a bit while at it, this should flatten it and keep the errors closer to their conditions.

```suggestion
  private elmToolingEntrypointsDecoder(json: unknown): NonEmptyArray<string> {
    if (typeof json !== "object" || json === null || Array.isArray(json)) {
      throw new Error(
        `Expected a JSON object but got: ${JSON.stringify(json)}`,
      );
    }

    if (!("entrypoints" in json)) {
      throw new Error(`There is no "entrypoints" field.`);
    }

    const { entrypoints } = json as { [key: string]: unknown };

    if (!Array.isArray(entrypoints) || entrypoints.length === 0) {
      throw new Error(
        `Expected "entrypoints" to be a non-empty array but got: ${JSON.stringify(
          json,
        )}`,
      );
    }

    const result: Array<string> = [];

    for (const [index, item] of entrypoints.entries()) {
      if (typeof item !== "string" || !item.startsWith("./")) {
        throw new Error(
          `Expected "entrypoints" to contain string paths starting with "./" but got: ${JSON.stringify(
            item,
          )} at index ${index}`,
        );
      }

      result.push(item);
    }

    return result as NonEmptyArray<string>;
  }
```

lydell (approved):
This looks good to me! I left one suggestion if you feel like it.

Razzeee:
Thanks, setting this to draft, so that I'm forced to test this properly :)

PR #505, opened 2021-02-07
Improve unused diagnostics performance

### LINKED ISSUE #504: UnusedPatternVariable diagnostic is very slow on big files
Continuation from #484.

When my project contains files like this ([Utilities.elm](https://ipfs.runfission.com/ipfs/bafybeid5qds2j2j2mfw6ndp5tz75uak2p2e6d3xs5s4ccghoekl3egvq3e/p/Utilities.elm)) it makes my fans spin indefinitely, because a process running `elm-language-client-vscode/server/out/index.js` runs 100% on one core.

I'm running the `elm-language-client-vscode` master (https://github.com/elm-tooling/elm-language-client-vscode/commit/7532d7fc0caf9a6590b15f817272337f09bc0704).

I've checked that tree-sitter-elm can parse this file and it's just fine.

Turning off the `ElmLS Diagnostics` setting fixes the problem (but is unfortunate, obviously).

I have created a [.cpuprofile](https://ipfs.runfission.com/ipfs/bafybeih5pyz6ibte2naqtcn2535jxlcpublw3gysnyucfxl2fvdz2uoeo4/p/CPU-20210205T115804.043Z.cpuprofile), but I doubt it's helpful. I *think* when I'm profiling a language server extension, I basically only profile the *client* part, but not the *server* part. This results in a profile that contains only native code, as 99% of the time its waiting for messages on sockets (this is my hypothesis). If you know any other way to profile this, I'd be happy to try it.

I would love input in how to further test this issue. Can anyone else reproduce this issue?
Maybe it's related to file size (and it's just running *reeeeally* long), as the file is almost 33kLOC, ~~but it has never run to completion, even after hours~~. This is incorrect! It finishes after ~10 minutes. It seems I got the impression it never finishes, because the process gets restarted once I save any other Elm file?

jmbockhorst:
This should greatly improve performance on large files. Fixes #504.

### DIFF
--- a/src/providers/completionProvider.ts
+++ b/src/providers/completionProvider.ts
@@ -370,7 +370,6 @@ export class CompletionProvider {
         ...this.getCompletionsFromOtherFile(
           checker,
           treeContainer,
-          params.textDocument.uri,
           replaceRange,
           targetWord,
         ),
@@ -526,7 +525,6 @@ export class CompletionProvider {
   private getCompletionsFromOtherFile(
     checker: TypeChecker,
     treeContainer: ITreeContainer,
-    uri: string,
     range: Range,
     inputText: string,
   ): CompletionItem[] {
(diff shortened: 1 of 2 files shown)


Razzeee:
LGTM.

Small note, I'm wondering, if we should show some kind of progress/work indicator, for linting.

matheus23:
Running this branch on the files I posted in #504 only takes ~50 seconds to check, instead of ~10 minutes, which is great :)

Razzeee:
@matheus23 would you be able to also run it through elm-analyse? so that we have a reference?

PR #506, opened 2021-02-08
Minor completion improvements

jmbockhorst:
Fixes for some things that have been bugging me.

- Fix union constructor auto imports
- Prefer `Int` over `Basics.Int` for Basics module
- No completions for lower patterns (param names)
- Record completions for record patterns
- Don't complete on import alias
- Improve case branch pattern completions

### DIFF
--- a/src/providers/completionProvider.ts
+++ b/src/providers/completionProvider.ts
@@ -142,6 +142,30 @@ export class CompletionProvider {
         // Don't complete in comments
         return [];
       } else if (
+        nodeAtPosition.parent?.type === "lower_pattern" ||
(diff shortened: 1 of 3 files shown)


jmbockhorst:
Probably should do another night of minor work and add some tests.

Razzeee:
Looks good so far

jmbockhorst:
Ok, I think this is ready now.

Razzeee:
The tests only cover the second commit, as far as I can tell. Will still merge, but kind of unexpected after you pointing out the above.

jmbockhorst:
I added tests for record patterns (from the first commit) and the rest are just sorting or not showing completions.

PR #507, opened 2021-02-09
Add code action to create missing record field

### LINKED ISSUE #477: Offer to create a record field if it is missing
This would be especially useful with the `Model` type, where it can be common to write out a record field before it exists. We should be able to infer the expected type of that field and create it.

jmbockhorst:
Closes #477.


Razzeee:
As we want this for the next minor version, I think at least one test would be nice. To be sure it's actually functional.

jmbockhorst:
Added tests and improved the infrastructure for testing code actions.

Issue #406, opened 2020-12-15
Tracking where diagnostics differ from the compiler


jmbockhorst:
Ongoing tracking issue for known areas where the diagnostics differ from the compiler.

### Places we don't show an error when there should be:
- [ ] Missing case branches 
- [ ] Unknown/missing exposed value in module exposing list
- [ ] Same value name imported from multiple modules. Ambiguous name

peret:
Not sure if I should raise a separate issue for this, but using reserved words as field-/argument-names also isn't detected by the language server. E.g. elm compiler gives:
```
I am partway through parsing a record type, but I got stuck on this field name:

80|     { name : Maybe String
81|     , type : LocationType
          ^^^^
It looks like you are trying to use `type` as a field name, but that is a
reserved word. Try using a different name!
```

but elm-language-server doesn't show any errors.

PR #521, opened 2021-02-16
Handle imports better and add import, exposing, and ambiguous diagnostics

### LINKED ISSUE #511: Diagnostics don't match compiler when shadowing Program
The Elm compiler seems to let you shadow certain types from the Prelude, like `List`, `Program`, etc.

`elm-ls` complains about this by noting you have the wrong number of type arguments. It's clearly not prioritizing the user-defined `Program`.
Attached is a repro, using `Program`; interestingly enough, it doesn't throw an error with `List`.

## Expected Behavior
The editor's diagnostics should match the compiler's, and show no errors.

## Current Behavior
You get an error from the editor indicating that `Program` has the wrong number of arguments.

Open the attached SSCCE.
[elm-ls-demo.zip](https://github.com/elm-tooling/elm-language-server/files/5954028/elm-ls-demo.zip)
Screenshot 2021-02-09 at 12 18 06@2x

## Context
In our case, these are the last remaining spurious errors in the codebase from the language server.

## Your Environment
* Version used: 2.0.3
* Editor used: Vim
* Environment name and version (e.g. node.js 5.4): node 12.18.1
* Operating System and version: MacOS 11.1

jmbockhorst:
Fixes #511 and part of #406. Lots going on here, but the main change is that `Imports` is rewritten to be more like how the compiler does it and so that we can produce errors for unknown imports. Unknown exposing diagnostics were added in the binder, and I took the time to refine the `findDefinition` api to produce ambiguous value/type diagnostics. I tried to add plenty of tests to cover all the cases. 

Fyi, those wildcard import tests are not valid in 0.19 anymore, so I removed them.


src/compiler/diagnostics.ts
```diff
@@ -61,6 +61,21 @@ function diag(
 }
 
 export const Diagnostics = {
+  AmbiguousType: diag(
+    "ambiguous_type",
+    "The usage of the type `{0}` is ambiguous because it is exposed by multiple imports.",
+    DiagnosticSeverity.Error,
+  ),
+  AmbiguousVar: diag(
+    "ambiguous_variant",
```
Razzeee, line 70: This should probably be `ambiguous_variable`?

test/diagnosticTests/elmDiagnostics.test.ts
```diff
@@ -832,4 +837,435 @@ test listList =
...
+module App exposing (..)
+
+type alias Program = String
+
+--@ Platform.elm
+module Platform exposing (..)
+
+type Program flags model msg = Program
+`;
+    await testTypeInference(basicsSources + stringSources + source2, []);
```
Razzeee, line 892: A bit weird, as the tests description says `produce an error`

Razzeee (changes requested):
LGTM. Just some small things.

test/diagnosticTests/elmDiagnostics.test.ts
```diff
@@ -838,7 +838,7 @@ test listList =
     await testTypeInference(basicsSources + source, []);
   });
 
-  test("test duplicate type imports produce an error", async () => {
+  test("test duplicate type imports produce an error if ambiguous", async () => {
```
Razzeee, line 841: This is also the case for the other tests that were added

Issue #484, opened 2021-01-24
Tree sitter parser get stuck in invalid state


jmbockhorst:
Sometimes the parser can get in an invalid state and and it gets stuck in an infinite loop. Not easily reproducible, but it seems to only happen on Windows for me.

Razzeee:
If it's really just windows, line endings are  immediately a suspect.

Anyway, it's going to be a bug in the c++ parser.

jmbockhorst:
It could still be a bug here. I've had this happen before when we use an old tree after a new one is parsed.

jmbockhorst:
Yes, it sounds similar.

matheus23:
I'm having issues with the language sever making my fans spin and using 100% of one core constantly.
As soon as I kill the process, a new one starts and makes my fans spin again :( I feel powerless here :smile: 

I'm on Arch Linux, though, so might not actually be related. Can I help you debug this somehow?
From a first glance, there's nothing suspicious in the (set to verbose) sever logs, but I'll investigate further.

Razzeee:
Did you test this with an open source project or the spa-example in the bast case?

matheus23:
> Did you test this with an open source project or the spa-example in the best case?

Nope! Not yet. But I'm pretty sure I can reproduce it with any elm codebase, I'll take a look!

I created a `.cpuprofile` and looked at it in the chrome dev tools profiler. 99% of the time is spent on `(program)`. According to google this means it's most likely native code. Which leads me to believe this has something to do with tree sitter parsing indeed.

I found this issue about infinite loops in tree sitter, maybe it's relevant/related?

https://github.com/tree-sitter/tree-sitter/issues/288

matheus23:
Oh! I **can't** reproduce this with `elm-spa-example`!

I have an idea, though: In almost all projects I'm generating elm code with https://github.com/monty5811/postcss-elm-tailwind or something similar.

And as far as I know, when you generate files with node, it includes line feeds!

Checking this hypothesis.

matheus23:
Hypothesis wrong. The file is `LF`, but it still causes my fans to spin.

matheus23:
Interestingly, this issue doesn't seem to be caused by tree-sitter.
I've cloned and run the master version of tree-sitter-elm and ran `npx tree-sitter parse Tailwind.elm` (`Tailwind.elm` being the file causing my fans to spin) and it doesn't cause an issue.

[...]

Issue #473, opened 2021-01-22
In version 2.0.0, returning lambda from the `case` block is reported as Parsing error.


shamansir:
Half of the body of the proper Elm file is reported as Parsing Error in version 2.0.0. The code in the file is marked as an error and marks as erroneous all the files which use this module. When I edit and save this file, the error disappears for some time, but then gets back again when I edit other files.

## Expected Behavior
The parsing error shouldn't be produced. Other proper files shouldn't be affected

## Current Behavior
This is considered as an error by the plugin:

```elm
flipBy : Flip -> Plane a -> Plane a
flipBy how (Plane (width, height) _ as plane) =
    plane |>
        transform
            (case how of
                Horizontal -> \(x, y) -> (width - 1 - x, y)
                Vertical -> \(x, y) -> (x, height - 1 - y))
```

While this isn't:

```elm
flipBy : Flip -> Plane a -> Plane a
flipBy how (Plane (width, height) _ as plane) =
    plane |>
        transform
            (\(x, y) -> (width - 1 - x, y))
```

## Steps to Reproduce (for bugs)
Source code, if would be needed: https://github.com/shamansir/kvant/blob/c8a09ed88c81c894832bb7012b6a4bc3187633b5/src/Kvant/Plane.elm

## Your Environment

* Version used: v2.0.0, Elm v0.19.1
* Editor name and version (e.g. VSCode 1.36.1): 1.52.1
* Environment name and version (e.g. node.js 5.4): node v12.16.1, 
* Operating System and version: macOS Big Sur, 1.11

jmbockhorst:
Running `elm-format` formats the case like this and the parsing error goes away:
```elm
(case how of
    Horizontal ->
        \( x, y ) -> ( width - 1 - x, y )

    Vertical ->
        \( x, y ) -> ( x, height - 1 - y )
)
```

Razzeee:
Yeah, expected that, just didn't have time to run elm format yet

Razzeee:
Problem is the `()` around case, we need the new line at the end of the case. So either putting it in the next line or removing the braces is what gets rid of this.

Razzeee:
https://github.com/elm-tooling/tree-sitter-elm/pull/74
Should help fix the error happening in the first place - but we still prefer you to use elm-format ;)

PR #528, opened 2021-02-21
Update dependencies and tree-sitter

Razzeee:
Fixes https://github.com/elm-tooling/elm-language-server/issues/473
Fixes https://github.com/elm-tooling/elm-language-server/issues/484


jmbockhorst (approved):
Looks good. I might look at exposing list completions, because I saw they weren't working the other day.

PR #532, opened 2021-02-23
Show completions for multiple modules if they share a name/alias

### LINKED ISSUE #405: Show completions from both modules when multiple imports have the same name/alias
When there are multiple imports with the same name or alias, the completions should show from both modules.

jmbockhorst:
Closes #405.

### DIFF
--- a/src/providers/astProvider.ts
+++ b/src/providers/astProvider.ts
@@ -143,7 +143,7 @@ export class ASTProvider {
     );
 
     oldNodes.forEach((node) => {
-      if (!newIds.has(node.id)) {
+      if (node && !newIds.has(node.id)) {
         params.program.getTypeCache().invalidateValueDeclaration(node);
         params.program.getTypeCache().invalidateTypeAnnotation(node);
         params.program.getTypeCache().invalidateTypeOrTypeAlias(node);
--- a/test/completionProvider.test.ts
+++ b/test/completionProvider.test.ts
@@ -1503,4 +1503,83 @@ func model =
 
     await testCompletions(source, ["details"]);
   });
+
+  it("Completions for module alias if using an import alias", async () => {
+    const source = `
+--@ Module.elm
+module Module exposing (..)
+
+foo = ""
+
+type Maybe a = Just a | Nothing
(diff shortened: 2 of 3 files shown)


src/providers/astProvider.ts
```diff
@@ -143,7 +143,7 @@ export class ASTProvider {
     );
 
     oldNodes.forEach((node) => {
-      if (!newIds.has(node.id)) {
+      if (node && !newIds.has(node.id)) {
```
jmbockhorst, line 146: Was getting null nodes here that we weren't before. Maybe from the tree-sitter upgrade.

Razzeee: Not sure how that's possible, it's defined in line 96. As `const oldNodes = tree?.rootNode.namedChildren ?? [];`.

test/completionProvider.test.ts
```diff
@@ -1503,4 +1503,83 @@ func model =
...
+
+import Module as M
+
+func =
+    M.f{-caret-}
+`;
+
+    await testCompletions(
+      source,
+      ["foo", "Maybe", "Just", "Nothing"],
```
Razzeee, line 1527: I think it's weird, that this case returns Maybe, Just etc as the caret is after the `f`

jmbockhorst: The client is responsible for filtering, this is how all our completions work.

src/providers/astProvider.ts
jmbockhorst: Tree sitter calls `delete` on some of those nodes?

Razzeee: Might be, anyway, this check doesn't hurt.

jmbockhorst: But it doesn't look like they do. All I know is that when it happened that array looked like `[Node, Node, ..., Node, null, null, null, ...]`

test/completionProvider.test.ts
Razzeee: Right, forgot about that.

PR #533, opened 2021-02-23
Handle case branch indentation better

jmbockhorst:


test/codeActionTests/addMissingCaseBranches.test.ts
```diff
@@ -0,0 +1,95 @@
...
+            ""
+
+        _ ->
+            Debug.todo "branch '_' not implemented"
+
+`;
+
+    await testCodeAction(
+      source,
+      [{ title: "Add missing case branches" }],
```
Razzeee, line 39: I think we should do another pass later on and let the user know beforehand, if we're going to add a catch all pattern

jmbockhorst: Yeah I was thinking we could do something better, but wasn't sure what.

Razzeee:
It would be nice to cover ```
module Test exposing (..)


func a =
    case a of
```

But that's a parsing error and might be hard to do

jmbockhorst:
That would be cool, but tricky to implement.

PR #537, opened 2021-02-28
Swap ListItems CodeAction

Strepto:
**What have i made?**

Related to https://github.com/elm-tooling/elm-language-client-vscode/issues/170

I have made a CodeAction to re-order lists. This action can move a list item up or down in the list, and can be invoked by keyboard shortcuts from your tool of choice.

moveactions

**What is it missing?**

- [ ] Handle comments
- [ ] Ensure handling of non-compiling code
- [ ] Add tests
- [ ] Refactoring
- [ ] (Non fixable?) Move cursor when swapping items, so it can be applied multiple times. (See: https://github.com/microsoft/language-server-protocol/issues/724 )


Strepto:
The current test is broken, and I cannot figure out how I should fix it. Would like a pointer to figure out how to handle the `undefined` at this line https://github.com/elm-tooling/elm-language-server/blob/032707ab925b841d6829140e7f87c3b7c80b9934/test/codeActionTests/codeActionTestBase.ts#L135

Razzeee:
Seems like we don't have tooling to test `registerRefactorAction` yet. The problem is, that you just let the server know which codeActions there are and the actual edit get's created at a later time. But our tests need to do that manually, which we don't do yet. So the call to create the edit instruction never happens right now.

Razzeee:
The new logic isn't perfect, as it should be separated some more, but should be fine for now.

src/providers/codeAction/swapListItemCodeAction.ts
```diff
@@ -0,0 +1,161 @@
...
+          refactorName,
+          uri: params.sourceFile.uri,
+          range: params.range,
+        },
+      });
+    }
+
+    if (canMoveNext) {
+      codeActions.push({
+        title: "Move List item Down",
```
Razzeee, line 48:
```suggestion
        title: "Move list item down",
```
I would expect these to be cased like this

src/providers/codeAction/swapListItemCodeAction.ts
```diff
@@ -0,0 +1,161 @@
...
+    );
+
+    const canMoveNext = getTargetNodes(nodeAtPosition, "next");
+    const canMovePrev = getTargetNodes(nodeAtPosition, "previous");
+
+    const codeActions: IRefactorCodeAction[] = [];
+
+    if (canMovePrev) {
+      codeActions.push({
+        title: "Move List item Up",
```
Razzeee, line 35:
```suggestion
        title: "Move list item up",
```
I would expect a casing like this

src/providers/codeAction/swapListItemCodeAction.ts
```diff
@@ -0,0 +1,161 @@
...
+import { PositionUtil } from "../../positionUtil";
+import { TreeUtils } from "../../util/treeUtils";
+import {
+  CodeActionProvider,
+  IRefactorCodeAction,
+  IRefactorEdit,
+} from "../codeActionProvider";
+import { ICodeActionParams } from "../paramsExtensions";
+
+// Handle adding annotation to let expr declaration
```
Razzeee, line 17: I think that's a c&p error

src/providers/codeAction/swapListItemCodeAction.ts
```diff
@@ -0,0 +1,161 @@
...
+function findSiblingSemanticListNode(
+  node: SyntaxNode,
+  direction: "previous" | "next",
+): SyntaxNode | null {
+  const isNext = direction == "next";
+
+  let target = iterate(node);
+  while (
+    (target?.type == "," ||
+      target?.type == "line_comment" ||
```
Razzeee, line 147: `block_comment` might be affected too

Strepto:
Thanks! I'll look into improving this when I have some time to spare.

Strepto:
Thanks for the comments! I'll try working on refactoring next, as I think the logic works better now after I have tested it. Also easier to refactor now that I have some Okish tests!

Razzeee:
> I would like to test some cases where a CodeAction should NOT be available. Such as "Move list item up" not being available on the first item. And in this particular test no Move list item should appear.)
> 
> Does this seem like something I/we could extend the testCodeAction code with @Razzeee ?

Should be possible, we already do this for similar tests (diagnostics), not sure if codeActionTests have the capability yet.

Strepto:
I have been monitoring https://github.com/microsoft/language-server-protocol/issues/724 for any updates but no news yet.

Without cursor controll I feel this does not yet feel as usable as it should be. 

I'll close this for now.  But I will reopen or create a new PR once the protocol has this support 🚀.

PR #542, opened 2021-03-04
Add command manager typing

Razzeee:
Not sure if that might be a problem in the future, but gets rid of a warning now.


jmbockhorst (changes requested):
Why don't we use `unknown`?

Razzeee:
I think we could, I tried that first, but `unknown` always needs you to unpack. The current code seems to be only using string anyway, so there shouldn't be much value. Might be if we add inputs with different types at some point.

PR #548, opened 2021-03-04
Fix pattern matching

jmbockhorst:


test/diagnosticTests/elmDiagnostics.test.ts
```diff
@@ -1424,6 +1425,88 @@ func result =
         _ ->
             ""
   `;
-    await testTypeInference(basicsSources + source, [], true);
+      await testTypeInference(basicsSources + source, [], true);
+    });
+
+    it("should not have an error with module prefixed ctor", async () => {
```
jmbockhorst, line 1431: @Razzeee I can't get a test to fail, but still seeing diagnostics in elm-spa-example. I know it's related to module prefixes (It fixes in elm-spa-example when you remove them), but don't know why I can't get the error here. If you have a chance to look at it and get a failing test that would be great.

jmbockhorst:
Finally found it. We should squash this.

Razzeee:
Yay 🎉

Issue #535, opened 2021-02-25
Add `Remove unused`


Razzeee:
Add remove unused action for 

- [x] functions
- [ ] types
- [ ] type alias

PR #549, opened 2021-03-05
More remove unused code actions

Razzeee:
Closes https://github.com/elm-tooling/elm-language-server/issues/535
Probably as far as we can take this type/type alias wise with the current diagnostics.


jmbockhorst:
This code looks good. 

Something I found is that we have a problem with overlapping ranges when using the "Remove all" code action. For example, lets say `Decoder` and `decodeString` are both unused, here are the ranges we calculate. Since they overlap, VSCode fails the edits.
```elm
                             --------
import Json.Decode exposing (Decoder, decodeString, field)
                                    --------------
```

I might look into this tonight.

jmbockhorst (approved):

PR #550, opened 2021-03-05
Handle removing all import values better

jmbockhorst:
Handle overlapping ranges when removing multiple import values at the same time. I also added special handling for the case when we remove all import values, it removes the exposing list. Still need to add tests and there is one case that still does not work.

- [x] When the last two (or more) values in a list are unused, an extra comma is left behind.
- [x] Add tests

### DIFF
--- a/src/providers/codeActionProvider.ts
+++ b/src/providers/codeActionProvider.ts
@@ -35,6 +35,7 @@ import { ExposeUnexposeHandler } from "./handlers/exposeUnexposeHandler";
 import { MoveRefactoringHandler } from "./handlers/moveRefactoringHandler";
 import { ICodeActionParams } from "./paramsExtensions";
 import { ElmPackageCache } from "../compiler/elmPackageCache";
+import { comparePosition } from "../positionUtil";
 
 export interface ICodeActionRegistration {
   errorCodes: string[];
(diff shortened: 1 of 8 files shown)


test/codeActionTests/codeActionTestBase.ts
```diff
@@ -91,6 +96,15 @@ export async function testCodeAction(
   workspaces.splice(0, workspaces.length);
   workspaces.push(program);
 
+  container.register(DiagnosticsProvider, {
+    useValue: new DiagnosticsProvider(),
+  });
+
+  // Needed be codeActionProvider uses these diagnostics
```
Razzeee, line 103: Typo

test/codeActionTests/removeUnusedCodeAction.test.ts
```diff
@@ -0,0 +1,223 @@
...
+`;
+
+    await testCodeAction(
+      source,
+      [{ title: "Remove unused value `Result`" }],
+      expectedSource,
+    );
+  });
+
+  it("removing an import at the end of exposing list", async () => {
```
Razzeee, line 85: Seems duplicated?

test/codeActionTests/removeUnusedCodeAction.test.ts
```diff
@@ -0,0 +1,223 @@
...
+
+    await testCodeAction(
+      source,
+      [{ title: "Remove all unused code" }],
+      expectedSource,
+      true,
+    );
+  });
+
+  it("removing all imports in exposing list except the end", async () => {
```
Razzeee, line 196: Maybe a test that keeps two values might be interesting here too?

src/providers/codeActionProvider.ts
```diff
@@ -221,6 +214,29 @@ export class CodeActionProvider {
...
+      // Handle if the ranges overlap
+      if (
+        lastEditEnd &&
+        newEditStart &&
+        comparePosition(lastEditEnd, newEditStart) > 0
+      ) {
+        edit.range.start = lastEditEnd;
+
+        if (comparePosition(edit.range.end, edit.range.start) < 0) {
+          edit.range.end = edit.range.start;
```
Razzeee, line 235: I'm not seeing, where sortedEdits/edit is used after this?

src/providers/codeActionProvider.ts
```diff
@@ -221,6 +214,29 @@ export class CodeActionProvider {
       }
     });
 
+    const sortedEdits = edits.sort((a, b) =>
```
Razzeee, line 217: Do you think we should filter for delete here? We shouldn't edit other action types right? We only happen to have delete action at the moment.

src/providers/codeActionProvider.ts
jmbockhorst: Yeah, that’s why I added the note above about mutation. The `edit` object that is updated is referenced in `edits`.

src/providers/codeActionProvider.ts
jmbockhorst: Any edits that overlap we would want to fix, because it would break on the client side. So I think we want this for all types.

PR #567, opened 2021-03-26
Explicitly set expose function code action to preferred

Razzeee:
Don't default to preferred = true for all codeActions

### DIFF
--- a/src/providers/codeAction/exposeUnexposeCodeAction.ts
+++ b/src/providers/codeAction/exposeUnexposeCodeAction.ts
@@ -39,6 +39,7 @@ CodeActionProvider.registerRefactorAction(refactorName, {
       } else {
         result.push({
           title: "Expose Function",
+          isPreferred: true,
           kind: CodeActionKind.Refactor,
           data: {
             actionName: "expose_function",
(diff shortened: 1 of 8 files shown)


jmbockhorst:
I went ahead and implemented a code actions preference system like VSCode Typescript has, so we can manage it better.

Razzeee:
Yeah, I actually agree, that should be the top choice in the future too

src/providers/codeAction/exposeUnexposeCodeAction.ts
```diff
@@ -39,6 +42,7 @@ CodeActionProvider.registerRefactorAction(refactorName, {
       } else {
         result.push({
           title: "Expose Function",
+          isPreferred: true,
```
Razzeee, line 45: This seems fishy to me. Should `IRefactorCodeAction` also start to extend ICodeAction? Should we omit `isPreferred` from that interface?

jmbockhorst: Yeah, that’s a good idea. I just forgot to remove it here.

Razzeee: Done

PR #569, opened 2021-03-31
Fix major performance issue with getting possible imports

jmbockhorst:
When `exp.node` was a `type_alias_declaration`, we were calling `findFirstNamedChildOfType` on the `file` many times, causing a big performance hit.


Razzeee:
Thank you

PR #579, opened 2021-04-25
Change default for singleFieldRecord rule to be false

Razzeee:
We now default to this to off, as we had multiple people point out that
they're no fans of this rule. And I think we agree, as we
just inherrited this from elm-analyze.


Razzeee:
Should hopefully be good to go now

jmbockhorst (approved):

PR #583, opened 2021-05-02
Find tests

frawa:
Experimenting.
The goal it to leverage the ELM Language Server to detect tests.

This will allow to reduce load time for the test explorer UI.
Currently, tests need to be run to be loaded into the UI.
This approach attempts a static code analysis approach to populate the test explorer UI.

The PR unblocks integration work on the client.

There are a few open questions to be addressed
- how to make the new request available earlier for clients? Or, how to hold back client calls until the request is available?
- is this the right way to use internal LS APIs? Does it leverage type inference enough?

### DIFF
--- a/src/compiler/references.ts
+++ b/src/compiler/references.ts
@@ -799,7 +799,7 @@ export class References {
       .map((field) =>
         TreeUtils.findFirstNamedChildOfType("lower_case_identifier", field),
       )
-      .filter(Utils.notUndefinedOrNull.bind(this))
+      .filter(Utils.notUndefinedOrNull)
       .filter((field) => field.text === fieldName);
   }
 
--- a/src/providers/findTestsProvider.ts
+++ b/src/providers/findTestsProvider.ts
@@ -85,6 +85,7 @@ export function findTestFunctionCall(
   node: SyntaxNode,
   typeChecker: TypeChecker,
 ): EFunctionCallExpr | undefined {
+  // console.log("FW0", typeChecker.typeToString(typeChecker.findType(node)));
   const call = findExpr("FunctionCallExpr", node);
   if (!call) {
     return undefined;
@@ -143,7 +144,10 @@ export function findTestSuite(
   if (!call) {
     return undefined;
   }
-  const label = findExpr("StringConstant", call.args[0])?.text;
+  const labelParts = findAllExprs("StringConstant", call.args[0])?.map(
+    (e) => e.text,
+  );
+  const label = labelParts?.length === 1 ? labelParts[0] : labelParts;
   if (label && isTestSuite(call, sourceFile, typeChecker)) {
     const testExprs = findExpr("ListExpr", call.args[1])?.exprList;
     const tests = testExprs
@@ -193,6 +197,23 @@ function findExpr<K extends keyof ExpressionNodeTypes>(
     node.type === type ? node : TreeUtils.findFirstNamedChildOfType(type, node);
   const e = mapSyntaxNodeToExpression(n);
   return e && mapExpr(key, e);
+}
+
+function findAllExprs<K extends keyof ExpressionNodeTypes>(
+  key: K,
+  node: SyntaxNode | undefined,
+): ExpressionNodeTypes[K][] | undefined {
+  if (!node) {
+    return undefined;
+  }
+  const type = typeByNodeType.get(key);
+  if (!type) {
+    return undefined;
+  }
+  const children = TreeUtils.findAllNamedChildrenOfType(type, node);
+  const ns = node.type === type ? [node, ...(children ?? [])] : children;
+  const es = ns?.map((n) => mapSyntaxNodeToExpression(n));
+  return es?.map((e) => e && mapExpr(key, e)).flatMap((v) => (v ? [v] : []));
 }
 
 function mapExpr<K extends keyof ExpressionNodeTypes>(
--- a/src/providers/findTestsProvider.ts
+++ b/src/providers/findTestsProvider.ts
@@ -33,17 +33,17 @@ export class FindTestsProvider {
     const connection = container.resolve<Connection>("Connection");
     connection.onRequest(FindTestsRequest, (params: IFindTestsParams) => {
       connection.console.info(
-        `Finding tests is requested ${params.workspaceRoot.toString()}`,
+        `Finding tests is requested ${params.projectFolder.toString()}`,
       );
       try {
         const elmWorkspaces: Program[] = container.resolve("ElmWorkspaces");
         const program = elmWorkspaces.find(
           (program) =>
-            program.getRootPath().toString() == params.workspaceRoot.toString(),
+            program.getRootPath().toString() == params.projectFolder.toString(),
         );
         if (!program) {
           // TODO dedicated error?
-          throw new NoWorkspaceContainsError(params.workspaceRoot);
+          throw new NoWorkspaceContainsError(params.projectFolder);
         }
         const typeChecker = program.getTypeChecker();
         const suites: TestSuite[] = Array.from(
@@ -84,7 +84,7 @@ export class FindTestsProvider {
         connection.console.info(
           `Found ${
             suites.length
-          } top test suites in ${params.workspaceRoot.toString()}`,
+          } top test suites in ${params.projectFolder.toString()}`,
         );
         return <IFindTestsResponse>{ suites };
       } catch (err) {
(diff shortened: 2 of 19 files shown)


src/providers/findTestsProvider.ts
```diff
@@ -0,0 +1,242 @@
...
+    });
+  }
+}
+
+// export for testing
+export function findTestFunctionCall(
+  node: SyntaxNode,
+  typeChecker: TypeChecker,
+): EFunctionCallExpr | undefined {
+  const call = findExpr("FunctionCallExpr", node);
```
jmbockhorst, line 83: This fails for `let_expr` because sometimes there are other functions that it finds first. We want to find that first function with the type or return type as checked below. Here is an example where it failed: https://github.com/rtfeldman/elm-css/blob/677a0ed2f43b5d480d379367b80dcf9688806090/tests/Keyframes.elm#L15

src/providers/findTestsProvider.ts
```diff
@@ -0,0 +1,242 @@
...
+    character: call.startPosition.column,
+  };
+  // TODO relative to workspace?
+  const file = sourceFile.uri.toString();
+  const label = labelParts?.length === 1 ? labelParts[0] : labelParts;
+  if (label && isTestSuite(call, sourceFile, typeChecker)) {
+    const testExprs = findExpr("ListExpr", call.args[1])?.exprList;
+    const tests = testExprs
+      ?.map((e) => findTestFunctionCall(e, typeChecker))
+      .map((call) => findTestSuite(call, sourceFile, typeChecker));
```
jmbockhorst, line 174: This can return a lot of `null` values if it can't find a test. Probably want to filter them out so it doesn't crash. For example, all of these will fail: https://github.com/rtfeldman/elm-css/blob/master/tests/Transitions.elm

jmbockhorst (commented):
Looks great so far! I did some testing using `elm-css` and found a few things.

src/providers/findTestsProvider.ts
Razzeee: If you decide to filter we have a utility that's used like this ` .filter(Utils.notUndefinedOrNull.bind(this))`

frawa: Perfect. There are more places to use that. 
Thanks.

Razzeee:
> how to make the new request available earlier for clients? Or, how to hold back client calls until the request is available? 

I don't think it can be made available earlier, as the language server first needs to process stuff. Holding back the client might be possible via `connection.onInitalized`, but you might need to extend our custom protocol for that. Not sure if there are any language server tools to do that.

src/providers/findTestsProvider.ts
frawa: Improved. Let's see if we can find more cases.

src/compiler/references.ts
```diff
@@ -799,7 +799,7 @@ export class References {
       .map((field) =>
         TreeUtils.findFirstNamedChildOfType("lower_case_identifier", field),
       )
-      .filter(Utils.notUndefinedOrNull.bind(this))
+      .filter(Utils.notUndefinedOrNull)
```
Razzeee, line 802: This might be a regression?

jmbockhorst: No, he changed the signature, so I think we don't need to bind anymore.

Razzeee: Heureka, just realized, what you did. Nevermind! :)

Razzeee:
@jmbockhorst 

I think we need index tests on startup too, basically run test detection/finding on init and save it to some collection. Then update it if something changes. Logic like this, should then just ask that collection.

This is more of a long term plan I guess, but should help us get rid of all the hard coded "tests" we have.

Algorithm to finding tests should be along these lines
https://github.com/elm-tooling/elm-tooling-cli/issues/17#issuecomment-834098781

test/findTestsProvider.test.ts
```diff
@@ -0,0 +1,383 @@
...
+module MyModule exposing (..)
+
+import Test exposing (..)
+
+topSuite = describe ("top suite" ++ "13") []
+`;
+
+    await testFindTests(source, [
+      {
+        label: undefined,
```
Razzeee, line 329: This test is not passing, after allowing `label` to be `undefined`. So I guess there might be something broken.

src/providers/findTestsProvider.ts
```diff
@@ -0,0 +1,277 @@
...
+    }
+  }
+  if (
+    t.nodeType === "Union" &&
+    t.module === "Test.Internal" &&
+    t.name === "Test"
+  ) {
+    return call;
+  }
+  // console.debug(
```
Razzeee, line 117: Let's remove this in a followup

src/providers/findTestsProvider.ts
```diff
@@ -0,0 +1,277 @@
...
+import { NoWorkspaceContainsError } from "../util/noWorkspaceContainsError";
+import { TreeUtils } from "../util/treeUtils";
+import { Utils } from "../util/utils";
+
+export class FindTestsProvider {
+  constructor() {
+    this.register();
+  }
+
+  private register(): void {
```
Razzeee, line 32: Let's move this into the constructor

src/providers/findTestsProvider.ts
```diff
@@ -0,0 +1,277 @@
...
+      }
+    });
+  }
+}
+
+// export for testing
+export function findAllTestSuites(program: IProgram): TestSuite[] {
+  const typeChecker = program.getTypeChecker();
+  return Array.from(program.getForest(true).treeMap.values())
+    .filter((sourceFile) => sourceFile.isTestFile)
```
Razzeee, line 68: This will ignore anything, that's outside of `tests`, but I think you know that and we should start out like this.

src/providers/findTestsProvider.ts
```diff
@@ -0,0 +1,277 @@
...
+
+function mapExpr<K extends keyof ExpressionNodeTypes>(
+  k: K,
+  e: Expression,
+): ExpressionNodeTypes[K] | undefined {
+  return e?.nodeType === k ? (e as ExpressionNodeTypes[K]) : undefined;
+}
+
+function stringLiteralToLabel(literal: string): string {
+  return String(JSON.parse(literal));
```
Razzeee, line 276: Can you help me to understand why we're doing this?

test/findTestsProvider.test.ts
```diff
@@ -0,0 +1,353 @@
...
+    await testFindTests(source, [
+      {
+        label: "top fuzz",
+        file: testModuleUri,
+        position: { line: 6, character: 6 },
+      },
+    ]);
+  });
+
+  // TODO improve inferred types?!
```
Razzeee, line 328: I think best would be to remove all todos and create issues for them. It's very hard for me to grasp, what your hinting at in these and I fear, that you won't remember too in some months :)

Razzeee: (I guess the skip is a little hint, still tracking this might be nicer)

Razzeee:
Merged this for now, but let's please talk about the comments I've raised :)

src/providers/findTestsProvider.ts
frawa: The labels reported for statically found tests need to fit the labels reported back from elm-tests json output, which are actual string values, no quotes, no escaping, just plain values.

The text found for expression of node type `StringConstant` are Elm-syntax string literals, which quotes and stuff.

The correct function to use would evaluate the literal in a way that complies to https://elm-lang.org/docs/syntax#literals, which does not seem to be very specific wrt escaping and special chars.
Do you have such a function already?

test/findTestsProvider.test.ts
frawa: Indeed, I could not figure it out and preferred to move up into client integration first.

Meanwhile, I am not sure it is a good idea to follow up on that route.
Given the know limitations between static analysis and runtime generated tests, solving this detail might not make a big difference.

src/providers/findTestsProvider.ts
frawa: I do not see how finding tests anywhere else would be useful.
Users will tell.

src/providers/findTestsProvider.ts
Razzeee: I don't think we already have that.

test/findTestsProvider.test.ts
Razzeee: Then we should probably get rid of this and the skipped test case.

src/providers/findTestsProvider.ts
frawa: See #591

test/findTestsProvider.test.ts
frawa: See #591

src/providers/findTestsProvider.ts
frawa: See #591

PR #587, opened 2021-05-04
Improve unused_pattern for empty constructs

Razzeee:


jmbockhorst:
Any test case or example of this?

Razzeee:
Added a test, which makes it painfully obvious, that the error message might be something we want to improve

jmbockhorst (approved):
Looks good.

PR #588, opened 2021-05-04
Add code action to create function in another module

### LINKED ISSUE #576: Add code action for creating missing function in other module
Similar to "Create local function", offer to create a function in another module.

jmbockhorst:
Closes #576.

### DIFF
--- a/src/compiler/typeChecker.ts
+++ b/src/compiler/typeChecker.ts
@@ -94,7 +94,7 @@ export interface TypeChecker {
   findImportModuleNameNode: (
     moduleNameOrAlias: string,
     sourceFile: ISourceFile,
-  ) => SyntaxNode | undefined;
+  ) => SyntaxNode[];
 }
 
 export function createTypeChecker(program: IProgram): TypeChecker {
@@ -549,7 +549,7 @@ export function createTypeChecker(program: IProgram): TypeChecker {
 
         const moduleNameOrAlias = nodeParent.text.substring(0, endPos);
         const moduleName =
-          findImportModuleNameNode(moduleNameOrAlias, sourceFile)?.text ??
+          findImportModuleNameNode(moduleNameOrAlias, sourceFile)[0]?.text ??
           moduleNameOrAlias;
 
         const moduleImport = findImport(sourceFile, moduleName, "Module")[0];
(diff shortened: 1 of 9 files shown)


src/compiler/typeChecker.ts
```diff
@@ -94,7 +94,7 @@ export interface TypeChecker {
   findImportModuleNameNode: (
     moduleNameOrAlias: string,
     sourceFile: ISourceFile,
-  ) => SyntaxNode | undefined;
+  ) => SyntaxNode[];
```
jmbockhorst, line 97: This changes because a import name or alias could return multiple imports. For example `App` would return both
```elm
import App
import Module.App as App
```

Razzeee: Should the name be updated to reflect what it returns?

jmbockhorst: Yeah, good point.

test/codeActionTests/codeActionTestBase.ts
```diff
@@ -141,15 +141,16 @@ export async function testCodeAction(
...
-      expect(
-        applyEditsToSource(
-          stripCommentLines(result.sources[uri]),
-          codeActions[testFixAll ? codeActions.length - 1 : 0].edit!.changes![
-            URI.file(baseUri + uri).toString()
-          ],
-        ),
-      ).toEqual(source);
+      const edits = changesToApply[URI.file(baseUri + uri).toString()];
+      if (edits) {
```
Razzeee, line 149: I'm wondering, if this should handle the case, where this is undefined or similar and just fail?

jmbockhorst: But a code action won't have edits for every file. Usually just one.

test/codeActionTests/makeExternalDeclarationFromUsageCodeActions.test.ts
```diff
@@ -0,0 +1,158 @@
...
+
+foo = ""
+`;
+
+    await testCodeAction(source, [
+      { title: `Create function in module 'Modules.App'` },
+      { title: `Create function in module 'App'` },
+    ]);
+  });
+});
```
Razzeee, line 158: Can we get a test case, for a case, where the file it's should be created in does not exist?

test/codeActionTests/codeActionTestBase.ts
Razzeee: Gotcha, I guess that would also mean, that we defined `expectedSources` wrong in the first place

jmbockhorst: I see what you mean. Yeah it would probably good to check all files.

PR #592, opened 2021-05-10
surround with module suite per sourcefile

frawa:
To conform with results from elm-test json reports, every module comes in a single test suite.
This makes matching suites from runs with suites from static load easier to match.

We want to match suites from those two sources to rescue as much location information as possible across edits and runs.


Razzeee:
Great work, thank you!

PR #593, opened 2021-05-11
Ignore empty test suites

Razzeee:
This behavior seems to match the behavior after running tests better.
Otherwise I end up with a lot of test suits in the
test explorer that never get run.

### DIFF
--- a/src/providers/findTestsProvider.ts
+++ b/src/providers/findTestsProvider.ts
@@ -77,8 +77,7 @@ export function findAllTestSuites(program: IProgram): TestSuite[] {
         .filter(Utils.notUndefined);
       return topSuites ? rootSuite(sourceFile, topSuites) : undefined;
     })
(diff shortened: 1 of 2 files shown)


src/providers/findTestsProvider.ts
```diff
@@ -77,7 +77,8 @@ export function findAllTestSuites(program: IProgram): TestSuite[] {
         .filter(Utils.notUndefined);
       return topSuites ? rootSuite(sourceFile, topSuites) : undefined;
     })
-    .filter(Utils.notUndefined);
+    .filter(Utils.notUndefined)
+    .filter((a) => a.tests && a.tests?.length > 0);
```
Razzeee, line 81: I'm unsure if this should be in the method or if it would be better to filter after it ran. If we keep it like this, we might want to rename it?

PR #594, opened 2021-05-11
Integrate with elm-review

w0rm:
elm-review

This is a super basic integration that supports diagnostics and code actions from `elm-review`. By default, `elm-review` is run if "review/src/ReviewConfig.elm" exists and if it is not disabled in the config.

The feedback from `elm-review` feels a little slow, we'll need to figure out how to integrate it in the watch mode, happy to try this in a separate PR.

@Razzeee thanks a lot for the initial pairing session!

### DIFF
--- a/src/providers/codeAction/elmReviewFixAction.ts
+++ b/src/providers/codeAction/elmReviewFixAction.ts
@@ -1,43 +1,20 @@
 import { TextEdit } from "vscode-languageserver";
-import { Utils } from "../../util/utils";
 import { CodeActionProvider } from "../codeActionProvider";
-import { IElmReviewDiagnostic } from "../diagnostics/elmReviewDiagnostics";
+import { hasElmReviewFixes } from "../diagnostics/elmReviewDiagnostics";
 
-const errorCodes = ["elm_review"];
-const fixId = "elm_review_fix";
 CodeActionProvider.registerCodeAction({
-  errorCodes,
-  fixId,
+  errorCodes: ["elm_review"],
+  fixId: "elm_review_fix",
   getCodeActions: (params) => {
-    return (<IElmReviewDiagnostic[]>params.context.diagnostics)
-      .map((diagnostic) => {
-        const { title, edits } = getEditsForDiagnostic(diagnostic);
-
-        if (title) {
-          return CodeActionProvider.getCodeAction(params, title, edits);
-        }
-      })
-      .filter(Utils.notUndefined);
+    return params.context.diagnostics
+      .filter(hasElmReviewFixes)
+      .map(({ data: { fixes } }) =>
+        CodeActionProvider.getCodeAction(
+          params,
+          "Apply elm-review fix",
+          fixes.map(({ range, string }) => TextEdit.replace(range, string)),
+        ),
+      );
   },
-  getFixAllCodeAction: () => {
-    return undefined;
-  },
+  getFixAllCodeAction: () => undefined,
 });
-
-function getEditsForDiagnostic(
-  diagnostic: IElmReviewDiagnostic,
-): { title?: string; edits: TextEdit[] } {
-  if (
-    diagnostic.data.code !== "elm_review" ||
-    diagnostic.data.fixes === undefined
-  ) {
-    return { edits: [] };
-  }
-
-  return {
-    title: "Apply elm-review fix",
-    edits: diagnostic.data.fixes.map((fix) =>
-      TextEdit.replace(fix.range, fix.string),
-    ),
-  };
-}
--- a/src/providers/codeAction/elmReviewFixAction.ts
+++ b/src/providers/codeAction/elmReviewFixAction.ts
@@ -1,0 +1,44 @@
+import { TextEdit } from "vscode-languageserver";
+import { Utils } from "../../util/utils";
+import { CodeActionProvider } from "../codeActionProvider";
+import { IDiagnostic } from "../diagnostics/diagnosticsProvider";
+
+const errorCodes = ["elm_review"];
+const fixId = "elm_review_fix";
+CodeActionProvider.registerCodeAction({
+  errorCodes,
+  fixId,
+  getCodeActions: (params) => {
+    return (<IDiagnostic[]>params.context.diagnostics)
+      .map((diagnostic) => {
+        const { title, edits } = getEditsForDiagnostic(diagnostic);
+
+        if (title) {
+          return CodeActionProvider.getCodeAction(params, title, edits);
+        }
+      })
+      .filter(Utils.notUndefined);
+  },
+  getFixAllCodeAction: () => {
+    return undefined;
+  },
+});
+
+function getEditsForDiagnostic(
+  diagnostic: IDiagnostic,
+): { title?: string; edits: TextEdit[] } {
+
+  if (
+    diagnostic.data.code !== "elm_review" ||
+    diagnostic.data.fixes === undefined
+  ) {
+    return { edits: [] };
+  }
+
+  return {
+    title: "Apply elm-review fix",
+    edits: diagnostic.data.fixes.map((fix) =>
+      TextEdit.replace(fix.range, fix.string),
+    ),
+  };
+}
--- a/src/providers/codeAction/elmReviewFixAction.ts
+++ b/src/providers/codeAction/elmReviewFixAction.ts
@@ -1,7 +1,7 @@
 import { TextEdit } from "vscode-languageserver";
 import { Utils } from "../../util/utils";
 import { CodeActionProvider } from "../codeActionProvider";
-import { IDiagnostic } from "../diagnostics/diagnosticsProvider";
+import { IElmReviewDiagnostic } from "../diagnostics/elmReviewDiagnostics";
 
 const errorCodes = ["elm_review"];
 const fixId = "elm_review_fix";
@@ -9,7 +9,7 @@ CodeActionProvider.registerCodeAction({
   errorCodes,
   fixId,
   getCodeActions: (params) => {
-    return (<IDiagnostic[]>params.context.diagnostics)
+    return (<IElmReviewDiagnostic[]>params.context.diagnostics)
       .map((diagnostic) => {
         const { title, edits } = getEditsForDiagnostic(diagnostic);
 
@@ -25,9 +25,8 @@ CodeActionProvider.registerCodeAction({
 });
 
 function getEditsForDiagnostic(
-  diagnostic: IDiagnostic,
+  diagnostic: IElmReviewDiagnostic,
 ): { title?: string; edits: TextEdit[] } {
-
   if (
     diagnostic.data.code !== "elm_review" ||
     diagnostic.data.fixes === undefined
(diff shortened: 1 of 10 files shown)


src/providers/diagnostics/diagnosticsProvider.ts
```diff
@@ -45,6 +47,10 @@ export interface IDiagnostic extends Omit<LspDiagnostic, "code"> {
   data: {
     uri: string;
     code: string;
+    fixes?: {
+      range: Range;
+      string: string;
+    }[]
```
w0rm, line 53: Not happy about this, but seems the only way to pass this information.

src/providers/diagnostics/elmReviewDiagnostics.ts
```diff
@@ -0,0 +1,167 @@
...
+              URI.parse(workspaceRootPath),
+              fileError.path,
+            ).toString();
+
+            const errors: IDiagnostic[] = fileError.errors.map(
+              (error: IError) => ({
+                message: error.message,
+                source: "elm-review",
+                range: toLsRange(error.region),
+                severity: DiagnosticSeverity.Warning,
```
Razzeee, line 146: if we have the data, we could set some tags here, I think only unused would be interesting:

`tags: DiagnosticTag.Unnecessary`

jmbockhorst:
I'm wondering if we should have a setting for the severity. Since elm-review doesn't have warnings, just errors that would fail a CI, I would want the option to see the diagnostics as errors. 

Maybe turn the bool setting into enum: `none`, `warning`, `error`?

src/providers/diagnostics/diagnosticsProvider.ts
jmbockhorst: What about instead adding something like this to `elmReviewDiagnostics.ts`?
```ts
export type IElmReviewDiagnostic = IDiagnostic & {
  data: {
    fixes?: {
      range: Range;
      string: string;
    }[];
  };
};
```

That way it is separated out into its own concern. And since you are already doing casting for the code actions, I think it works fine.

src/providers/diagnostics/elmReviewDiagnostics.ts
w0rm: Interesting! We should be able to derive this from the rule name, if it starts with `NoUnused.`

jfmengels: I think you can do it if it starts with `NoUnused`, not including the `.`.

src/providers/diagnostics/elmReviewDiagnostics.ts
```diff
@@ -0,0 +1,167 @@
...
+    if (
+      settings.disableElmReviewDiagnostics ||
+      !existsSync(path.join(workspaceRootPath, "review", "src", "ReviewConfig.elm"))
+    ) {
+      return fileErrors;
+    }
+
+    const elmReviewCommand: string = settings.elmReviewPath;
+    const options = {
+      cmdArguments: ["--report", "json"],
```
jfmengels, line 100: Could you add `--namespace=vscode`? This would help avoid having several processes (like this integration and an `elm-review --watch` running in a shell) from attempting to read/write to the same file at the same time.

I've described this in https://github.com/jfmengels/node-elm-review/blob/master/documentation/tooling-integration.md#namespacing

By the way, if the language server somehow knows about the paths to `elm` and `elm-format`, running with `--compiler=<path-to-compiler>` and `--elm-format-path <path-to-elm-format>` could maybe avoid problems where those aren't found.

src/providers/codeAction/elmReviewFixAction.ts
```diff
@@ -0,0 +1,44 @@
...
+    diagnostic.data.code !== "elm_review" ||
+    diagnostic.data.fixes === undefined
+  ) {
+    return { edits: [] };
+  }
+
+  return {
+    title: "Apply elm-review fix",
+    edits: diagnostic.data.fixes.map((fix) =>
+      TextEdit.replace(fix.range, fix.string),
```
jfmengels, line 41: A few things here about how `elm-review` applies a fix.

1. It applies fixes in the [reverse order of the range](https://github.com/jfmengels/elm-review/blob/master/src/Review/Fix.elm#L239), because if you start with the first one, then the ones that come after may have moved, producing unexpected results.

2. It does a few checks before applying a fix which you will have to do again here, that make sure that the fix is valid and won't result in bad Elm code, or makes no sense. The code for that is [here](https://github.com/jfmengels/elm-review/blob/master/src/Review/Fix.elm#L205-L208). At the moment (since a recent version), `elm-review`, when these checks find something weird, we give a big warning telling to contact the rule author, and we then ignore the fix. 
	- [Check if there are collisions](https://github.com/jfmengels/elm-review/blob/master/src/Review/Fix.elm#L231). (simplified) Having 2 fixes at `{ start: 1, end: 5 }` and `{ start: 3, end: 7 }` make it confusing which one to apply first.
	- Check that the source code before and after are different. If you apply no changes, then something went wrong probably
	- Check that the source code is still valid Elm code, which we do by parsing the file.

The reason why we don't do these checks when not running with `--fix`, is because it can take a significant amount of processing time to check whether each fix is valid or not, since that means actually applying each fix and parsing the resulting file. Therefore we only do it when necessary.

jmbockhorst: The language client and VS Code should take care of applying the text edits and adjusting ranges. It also handles overlapping ranges. The only thing missing would be checking if the source would be valid, but I don't think that is necessary.

src/providers/codeAction/elmReviewFixAction.ts
```diff
@@ -0,0 +1,44 @@
...
+        const { title, edits } = getEditsForDiagnostic(diagnostic);
+
+        if (title) {
+          return CodeActionProvider.getCodeAction(params, title, edits);
+        }
+      })
+      .filter(Utils.notUndefined);
+  },
+  getFixAllCodeAction: () => {
+    return undefined;
```
jmbockhorst, line 23: Can you implement a fix all code action?

jfmengels: I don't think this would be a good idea, at least not with a naive approach.

The given fixes for a single run will only make sense for that version of the project.

Here's an example of where it would not work. Imagine in this example, both `a` and `b` are unused, and the module is never referenced.

```elm
import M exposing (a, b)
```

The fix for removing `a` in this case would be removing the range corresponding to `a, `, and for `b` it would be removing the range corresponding to `, b`. If you try to apply both of these naively, you might get conflicting fixes because the ranges overlap. At worst you'll end up with `import M exposing (` and at best with `import M exposing ()`.


The way `elm-review` currently does fix-all is by doing the following logic:
- Review the code
- while there are valid fixes:
  - Take the first fix and apply it (in memory)
  - Review the code
- Present the user with the complete fix (for multiple files) which they can accept or not*

When done that way, the fix for `a` will be applied first, resulting in `import M exposing (b)`, and the second fix will remove the import entirely.

I don't know if it would be a better idea, but I imagine something like running `elm-review` in watch mode, applying the first fix, wait until `elm-review` notices the change and reports new things, and so on until either there are no more fixes, or that the user changes something in their project. All the while, showing that there is an ongoing process through a loader or something somewhere.

It would not be very fast, and it isn't very fast either in `elm-review`, but at least it is very reliable and it will find more things than a naive fix-all.


*I don't know how showing what the suggested fix could be done best in the context of an editor. It would be nice to show what the fix(es) will do before applying it I'd say? Even if you can do "Undo", if you don't see that something changed, then the user might get surprises?

jmbockhorst: Yeah, for our unused code actions I had to add extra logic to handle these cases: https://github.com/elm-tooling/elm-language-server/blob/19c33470138eaf8643f98131ceb912c504b05fd3/src/providers/codeActionLs/removeUnusedCodeAction.ts#L219.

We could do a similar thing for the specific cases where there would be overlap, but it would be tricky. What are the fixes that you know of that could overlap?
- Exposing lists
- Type constructors
- ??

jfmengels: Since rules can be custom-made, it can be anything and everything unfortunately 😬

For instance, if you have `a = if True then 1 else 2`< the `Simplify` rule will change it to `a = 1` but if it's also unused then `NoUnused.Variables` will remove it entirely.

In short, I don't think it's possible (or reasonable) to try to handle these overlaps. 😕

jmbockhorst: Yeah, I think it makes sense to not have a fix all.

src/providers/diagnostics/elmReviewDiagnostics.ts
Razzeee: are you doing a find up in node_modules if they don't get passed? if that's the case, we only need to provide the paths, if they are explicitly set in our settings.

everything else might be hard, as we rely heavily on `execa` to figure that out at the moment.

jfmengels: If they get passed, it doesn't try a fallback.
Otherwise, it tries running `elm` globally and running `npx elm`.
If you are sure that the executables are there, I think it's good to pass them, otherwise just skip it. I have not heard about people having problems with this :man_shrugging:

Razzeee: It's auto configuration, until you configure it. So that it's easy to use for anybody.

src/providers/codeAction/elmReviewFixAction.ts
w0rm: Good to know that we're getting this for free from VS Code 👍

src/providers/diagnostics/elmReviewDiagnostics.ts
w0rm: `npx elm` should cover the most of cases. I say let's wait for any user complaints and see if we actually need this.

w0rm:
@jmbockhorst @Razzeee I addressed the comments, could you take a final look?

src/providers/diagnostics/elmReviewDiagnostics.ts
Razzeee: I'll add a commit for cases where the paths were explicitly set

Razzeee:
Changes look great, thanks for doing this @w0rm 

Some todos for us after this

- [ ] Double check documentation
- [ ] Update client

Razzeee:
Client update:
https://github.com/elm-tooling/elm-language-client-vscode/pull/188/files

Razzeee (approved):
LGTM. Should be ready for merging.

jmbockhorst:
Can we do a similar thing that we do with ElmLs diagnostics, where if the configuration changes, we turn on/off the diagnostics dynamically? I can also do it after we merge this if you don't want to.

w0rm:
@jmbockhorst it runs on each save, and every time it checks is the setting is `off` — then it returns an empty `Map` that clears all the existing diagnostics. Given that we decided to ship it `off` by default, and hopefully try to integrate in the `watch` mode, I wonder how useful is this requirement.

jmbockhorst:
Oh yeah. Then it should be fine.

w0rm:
@jmbockhorst @Razzeee  thanks for the review!

Issue #589, opened 2021-05-06
Incorrect "Could not find module to import named ..." error on long file path


bwells:
I'm receiving an incorrect import error diagnostic `Could not find module to import named 'Request.AssignedTests.CreateAssignedTest' in 'dependencies' or 'source-directories'`

## Expected Behavior
`elm-language-server` reports no import error. 
`elm-language-server` warning matches `elm make` output.

## Current Behavior
elm-language-server reports an import error that does not match CLI `elm make` usage. I am able to build this code via `elm make` with zero errors or warnings. Additionally, other first party code within this source structure is imported without warning. 

## Possible Solution
This seems to be related to the file path length of the import. If I refactor the module from `Request.AssignedTests.CreateAssignedTests` to `Request.Asts.Cats` the import error is no longer reported.

## Steps to Reproduce (for bugs)
1. clone https://github.com/bwells/els-error
2. view `src/Main.elm` to observe import warning error
3. `elm make src/Main.elm` to confirm that the import warning is incorrect

## Context
This is an annoyance, in that I must navigate around incorrect warnings, but I am still able to use elm-language-server in day to day work. 

## Your Environment
* Version used: elm-language-server 2.1.0
* Editor used: Neovim nightly with included LSP client and ALE LSP client.
* Environment name and version (e.g. node.js 5.4): node 16.0.0
* Operating System and version: MacOS 11.3 on ARM

Thanks in advance for the help and for the work that has already gone into elm-language server.

Razzeee:
Can confirm you example to reproduce, even with the lastest version.

Razzeee:
The problem is the path of the file, which ends up as a test dependency due to it having `tests` in the path.

https://github.com/elm-tooling/elm-language-server/blob/main/src/compiler/program.ts#L628

bwells:
Oof. Well I work for a company that offers tests, as in exams, so getting rid of the word "test" in our codebase definitely isn't going to be feasible.

Razzeee:
Agreed, our test handling isn't as nice as it could/should be.
There are some other tickets already tracking it https://github.com/elm-tooling/elm-language-server/issues/497 for e.g.

bwells:
I'm able to run a custom build with a one line patch to essentially disable test detection support, so that has me sorted. Improved test handling would certainly be nice, but my two cents is it's not implementing any sort of interim solution on my account.

Razzeee:
[...]

PR #597, opened 2021-05-12
Align test detection logic

Razzeee:
Closes https://github.com/elm-tooling/elm-language-server/issues/589

This aligns the one place where we detect tests differently. We now basically only handle the `tests` folder.

### DIFF
--- a/src/compiler/program.ts
+++ b/src/compiler/program.ts
@@ -622,11 +622,10 @@ export class Program implements IProgram {
 
       const moduleName = utils.getModuleName(matchingPath, sourceDir);
 
-      // We could track this using the separate `testDirectories`
-      // But those are just hardcoded to ./tests anyways, so this should be fine
-      const isTestDir =
-        matchingPath.includes("tests") || matchingPath.includes("Tests");
-      if (isTestDir) {
+      const isTestFile =
+        this.getSourceDirectoryOfFile(matchingPath)?.endsWith("tests") ?? false;
+
+      if (isTestFile) {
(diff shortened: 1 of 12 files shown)


src/compiler/program.ts
```diff
@@ -622,11 +622,10 @@ export class Program implements IProgram {
 
       const moduleName = utils.getModuleName(matchingPath, sourceDir);
 
-      // We could track this using the separate `testDirectories`
-      // But those are just hardcoded to ./tests anyways, so this should be fine
-      const isTestDir =
-        matchingPath.includes("tests") || matchingPath.includes("Tests");
-      if (isTestDir) {
+      const isTestFile =
+        this.getSourceDirectoryOfFile(matchingPath)?.endsWith("tests") ?? false;
```
jmbockhorst, line 626: This feels weird to me. It's going to find less test directories in this case. 

I think ideally we wait until we parse the file, and if it contains a function with a `Test` return type, we know its a test file. And theoretically we would want to check this on every file change, but I can't think of a good performant way to do that.

Razzeee: Correct, it will find less, but it will align with all the other logic we have. And if we can make all other places depend on this central piece, we can maybe later do the correct algorithm. Which would be everything in `tests` + every file outside of tests, that has a `Test` return type.

https://github.com/elm-tooling/elm-tooling-cli/issues/17#issuecomment-834098781 is only partially correct, as they don't have the concept of a `TestFile`

Issue #603, opened 2021-05-31
elm-review diagnostics disappear when opening file from Problems panel


edkelly303:
## Expected Behavior
When I click on a diagnostic from elm-review in the Problems panel, it should take me to the line of code that the diagnostic refers to, and I should be able to see a yellow squiggly line under the problematic section of code. The yellow squiggly line should not disappear until I have fixed the problem.

In the VS Code Problems tab, I can see lots of elm-review warnings, and when I click on one of them to visit the relevant Elm file, I see the yellow squigglies for about a second; then they disappear. I think this happens when the code lenses load, and it also seems to trigger a refresh of everything in the Problems panel. If I then manually save the Elm file, then the elm-review squigglies reappear.

## Current Behavior
When I click on a diagnostic from elm-review in the Problems panel, it takes me to the line of code that the diagnostic refers to, and I can see a yellow squiggly line under the problematic section of code. 

However, after about a second, the yellow squiggly line disappears, as do all other diagnostics from elm-review in the file. The file also seems to disappear from the Problems panel, although other files with elm-review diagnostics are still present.

The disappearance seems to coincide with the loading of the code lenses - there is a moment when the vertical alignment of the code shifts (because the code lenses are being added to the file), and that's when the disappearance happens.

I can get the diagnostics to reappear in both the file and the Problems panel by manually saving the file (e.g. Ctrl+s), and if I keep resaving without making any changes, the squigglies persist. However, if I then make a small change (such as inserting a space at the end of a line) and re-save, the diagnostics disappear again. So it may be related to elm-format?

## Steps to Reproduce (for bugs)
See current behaviour section above

## Your Environment
* Version used: v2.2.0-RC1
* VSCode 1.56.2 
* NodeJS 14.16.0
* Xubuntu 20.10
* elm-format 0.85
* elm-review 2.4.5
* elm-test 0.19.1-revision7

PR #604, opened 2021-06-07
Fix elm-review diagnostics getting removed too eager

Razzeee:
Closes https://github.com/elm-tooling/elm-language-server/issues/603

@w0rm FYI

@jmbockhorst Would you consider this correct or can this be handled correctly inside of the `immediate` somehow?

### DIFF
--- a/src/providers/diagnostics/diagnosticsProvider.ts
+++ b/src/providers/diagnostics/diagnosticsProvider.ts
@@ -3,9 +3,9 @@ import {
   CancellationToken,
   Connection,
   Diagnostic as LspDiagnostic,
-  Range,
   DiagnosticSeverity,
   FileChangeType,
+  DidChangeTextDocumentParams,
 } from "vscode-languageserver";
 import { TextDocument } from "vscode-languageserver-textdocument";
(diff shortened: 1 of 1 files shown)


w0rm (approved):

src/providers/diagnostics/diagnosticsProvider.ts
```diff
@@ -218,9 +219,15 @@ export class DiagnosticsProvider {
       }
     });
 
-    this.documentEvents.on("change", () => {
+    this.documentEvents.on("change", (params: DidChangeTextDocumentParams) => {
       this.change();
 
+      this.updateDiagnostics(
```
jmbockhorst, line 225: Why do we need this in the first place? Don't they just get reset when we get new elm-review diagnostics? I wouldn't expect them to go away just because I make a change in a different part of the file.

w0rm: Two reasons: 
1. elm-review assumes the code passes `elm make`, any change to files can invalidate this
2. code actions, coming from elm-review fixes, contain locations that will be off after the file is edited

PR #609, opened 2021-06-17
Fix false positive for unnecessary list concat diagnostics

### LINKED ISSUE #605: False positive List.concat hint
image
```elm
test =
    let
        aList =
            if True then
                [ 2, 3 ]

            else
                []
    in
    List.concat
        [ [ 1 ]
        , aList
        , [ 4 ]
        ]
```

## Expected Behavior

No warnings.

## Current Behavior

ElmLS gives me a warning on the `List.concat` expression:

> You should just merge the arguments of `List.concat` to a single list. - ElmLS

## Possible Solution

don't show warnings in these cases. (Optimize for having false negatives sometimes, instead of having false positives sometimes?)

1. Paste the code above into e.g. VSCode and wait for ElmLS to scan the file & report warnings

## Your Environment

* Version used: 2.1.0
* Editor used: VSCode 1.56

jmbockhorst:
Fixes #605. I couldn't find a way to fix it with the query, so I just added a filter.

### DIFF
--- a/src/providers/diagnostics/elmLsDiagnostics.ts
+++ b/src/providers/diagnostics/elmLsDiagnostics.ts
@@ -287,7 +287,7 @@ export class ElmLsDiagnostics {
               ("," . (list_expr))*
               .
               "]"
-            )
+            ) @listExpr
           ) @functionCall
           (#eq? @target "List.concat")
         )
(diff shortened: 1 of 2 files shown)


Razzeee:
Looks great, I also couldn't figure out how to do it with the query directly (and without breaking anything else).

Thank you.

PR #614, opened 2021-07-11
Add check config for "Add type annotation" diagnostic

Razzeee:


src/providers/diagnostics/diagnosticsProvider.ts
```diff
@@ -453,7 +459,23 @@ export class DiagnosticsProvider {
                         sourceFile,
                         serverCancellationToken,
                       )
-                      .map(convertFromCompilerDiagnostic),
+                      .map(convertFromCompilerDiagnostic)
+                      .filter((diagnostic) => {
+                        const elmAnalyseJson = this.elmAnalyseJsonService.getElmAnalyseJson(
```
jmbockhorst, line 464: Can this be moved out of here so we aren't calling it for every diagnostic in the list?

Razzeee: Good catch, hopefully better like this.

PR #629, opened 2021-09-04
Remove the need for "entrypoints" in elm-tooling.json

### LINKED ISSUE #623: Consider getting rid of "entrypoints" in elm-tooling.json
Getting rid of "entrypoints" in elm-tooling.json would be nice because:

- Less configuration is always nice!
- Keeping "entrypoints" in elm-tooling.json can be a bit tedious in large projects with more and more entrypoints.
- It's not obvious what to put there for packages.

As far as I understand, "entrypoints" is used to call `elm make $entrypoints --report json --output /dev/null` instead of `elm make $current_file --report json --output /dev/null`. This gives you diagnostics for the entire project instead of just the current file (and its imports?).

https://github.com/elm-tooling/elm-language-server/blob/eb912794a79e4a786bca2f3ec8dcb0706fbeecaa/src/providers/diagnostics/elmMakeDiagnostics.ts#L200-L228

The idea is to run `elm make $all_elm_files_in_the_project --report json --output /dev/null` instead. I do that regularly. Something like this: `git ls-files | grep '\.elm$' | grep -v tests | grep -v review | xargs elm make --output /dev/null`. It’s fast!

Then the issue becomes mostly about finding all Elm files. I’m a bit fuzzy on how this works in a language server, or even how VSCode defines stuff like a “project” or how multiple elm.json are handled. Here are some notes:

- One `elm make` call per elm.json is needed. But that should be the case already?
- We don’t want to venture into `node_modules/`, `elm-stuff/` or `some-folder-of-generated-code/` that the user has .gitignore:d.
- Test files cannot use `elm make`, they must use `elm-test make`. But I already saw machinery to handle test files – can that be reused with this approach?
- My example command above with `xargs` filters away stuff in the `review/` folder (from elm-review). But there’s nothing special with elm-review, you just need a separate `elm make` call since it has its own elm.json: `cd review && git ls-files | grep '\.elm$' | grep -v tests | xargs elm make --output /dev/null`
- For package elm.json:s, `elm make $exposed_modules` should do it, right?

To me, this sounds doable! And I think it can be performant, too. What do you think?

lydell:
Fixes #623 

See https://github.com/elm-tooling/elm-language-server/pull/629#issuecomment-922380423 and the comment after that for a more up-to-date description of the approach, and some performance numbers.

Original PR description (outdated):

This is a quick proof of concept. If you think this looks good, I’ll continue and finish up this PR.

Running `elm make --output /dev/null src/Every.elm src/Single/Elm/File.elm src/You/Have.elm` is super fast, so there shouldn’t be a need to know the entrypoints. We can just compile _everything!_ Benefits:

- No configuration needed.
- Simpler code. We already know all Elm files anyway.
- This supports compiling files which aren’t imported from any entrypoint yet.

Performance – tested on the ~160k lines of Elm code at work:

- Just entrypoints: ~150 ms
- All files: ~220 ms

I’ve tested this a little bit with elm-spa-example and a package, and it seems to work! But I’m all new to both the language server and to VSCode extension development, so I’d appreciate help with:

- pointers on how to test more thoroughly
- help with testing :)
- should I write a test somewhere?

---

Cleanup todos when all is good to go:

- [x] Remove code that has to do with elm-tooling.json
- [x] Remove docs about elm-tooling.json (maybe replace it with some text saying that we _used_ to have a file called elm-tooling.json but you don’t need it anymore if you only have "entrypoints" in there).
- [x] Same docs update in https://github.com/elm-tooling/elm-language-client-vscode


src/providers/diagnostics/elmMakeDiagnostics.ts
```diff
@@ -78,19 +78,15 @@ export class ElmMakeDiagnostics {
...
-      (issues) => {
-        return issues.length === 0
-          ? new Map([[filePath.toString(), []]])
-          : ElmDiagnosticsHelper.issuesToDiagnosticMap(
-              issues,
-              workspaceRootPath,
-            );
-      },
-    );
+    const program = this.elmWorkspaceMatcher.getProgramFor(filePath);
```
lydell, line 81: It feels wrong to me to assume that all files belong to exactly one program (is that one elm.json?). The same source file can be part of multiple elm.json/projects, but I guess that’s rare? And out of scope for this PR? :thinking:

src/providers/diagnostics/elmMakeDiagnostics.ts
```diff
@@ -192,56 +188,85 @@ export class ElmMakeDiagnostics {
...
+    const workspaceRootPath = program.getRootPath().fsPath;
+
+    const fileToRelativePath = (file: ISourceFile): string =>
+      path.relative(workspaceRootPath, URI.parse(file.uri).fsPath);
 
-    this.connection.console.info(
-      `Find entrypoints: ${message}. See https://github.com/elm-tooling/elm-language-server#configuration for more information.`,
+    // Exclude installed packages (dependencies).
+    const isProjectPath = (pathString: string): boolean =>
+      !pathString.startsWith("..") && !path.isAbsolute(pathString);
```
lydell, line 203: Maybe not the nicest check but it seems to do the trick...

lydell: There’s one case where this _won’t_ work: If you set `ELM_HOME` to be _inside_ the workspace directory (such as inside `elm-stuff/`). So we might need a better way to tell dependencies apart from actual project files.

src/providers/diagnostics/elmMakeDiagnostics.ts
razzeee: Well, we can only show one state at a time in the language server/ your IDE. So we only run on the top most elm.json(s) we find.

If we had multiple programs for one filePath find all references will be confused and would first need to ask you, for which program you would like it to run.

razzeee:
Unfortunatly this part of the codebase has no tests, as I mostly inherited it and I find it quiet hard to test code, that calls out via shell.

It might be interesting, to see what now happens, with files, that are not referenced in Main.elm. They should have been ignored before, as we correctly walked the dependency tree. Now they might cause diagnostics to show up.

I guess that's a part of the tradeoff we could decide between for this.

- No config
	- Slower
	- Can cause misleading diagnostics, for files that are not referenced

- Entrypoint config
	- Needs setup and maintendence

lydell:
I expect any Elm file I have open to get diagnostics. For me, it’s pretty common to create a new Elm module (file), work on it for a while and then start using it in my app. During that initial work diagnostics will be helpful, won’t they?

> Can cause misleading diagnostics, for files that are not referenced

Can you expand on this? How can we get misleading diagnostics? I often do what the code in this PR does manually on the command line:

```bash
git ls-files | grep '\.elm$' | grep -v tests | grep -v review | xargs elm make --output /dev/null
```

That’s a nice way of knowing that the entire project compiles, and I’ve never seen any misleading output from it.

razzeee:
I think your right, I mixed up workspace handling with compiling. I think it should be sane as it only handles this part.

But I think we can end up getting errors for files that are not in `main` code path. I think that's a good thing, but know that it's something people have been doing. As parsing all elm code showed me, some packages from the package repo still have tests for 0.18 which might not compile.

lydell:
> some packages from the package repo still have tests for 0.18 which might not compile.

Isn’t it correct to show errors for those tests then? It should nudge the authors to migrate the tests to 0.19, or remove them.

razzeee:
> Isn’t it correct to show errors for those tests then? It should nudge the authors to migrate the tests to 0.19, or remove them.

Yes, that's what i've tried to say :)

I'm curious if this will affect people that have test files outside of the `tests` folder. As our detection is not correct at the moment and only cares for that folder.
So we might end up compiling those tests with elm-make and they might error due to that (missing deps etc)

Needs to be tested I think.

lydell:
Ah good point about tests outside `tests/`. I’ll try that out. If problematic, better test detection would be good anyway. Idea: If a file contains `import Test` it’s a test.

razzeee:
Sounds better, however, I'm not familiar with big test suits, so unsure, if people have helper modules that don't import `Test`

lydell:
A thing I planned on doing was to run `elm make` and `elm-test make` in parallel. But then I saw this PR: https://github.com/elm-tooling/elm-language-server/pull/419/files. “Make binary execution sync”. But it doesn’t say why. I would need async calls in order to parallelize.

razzeee:
I'm not sure anymore, I might have a hunch, but it might be wrong.

package.json
```diff
@@ -9,7 +9,7 @@
     "out"
   ],
   "engines": {
-    "node": ">=12.0.0"
+    "node": ">=12.9.0"
```
lydell, line 12: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled#browser_compatibility

src/compiler/utils/elmUtils.ts
```diff
@@ -5,16 +5,14 @@ import { URI } from "vscode-uri";
...
 export const isWindows = process.platform === "win32";
 
 /** Options for execCmdSync */
-export interface IExecCmdOptions {
+export interface IExecCmdSyncOptions {
   /** Any arguments */
   cmdArguments?: string[];
-  /** Shows a message if an error occurs (in particular the command not being */
-  /* found), instead of rejecting. If this happens, the promise never resolves */
-  showMessageOnError?: boolean;
```
lydell, line 17: This was unused

src/providers/diagnostics/elmMakeDiagnostics.ts
```diff
@@ -192,63 +188,165 @@ export class ElmMakeDiagnostics {
...
+    //   because of:
+    //   - The project doesn’t use `main =`, like `review/` for elm-review.
+    //   - The user has accidentally remove `main =` or not exposed it.
+    //
+    // Since these _could_ be test, we compile them with `elm-test make` rather
+    // than `elm make`, so that "test-dependencies" are allowed. If they _aren’t_
+    // tests, the only downside of this is that if you accidentally import a
+    // test-dependency, you won’t get an error for that. It should be an OK tradeoff.
+    const possiblyTestFiles = otherFiles.filter(
+      (file) => !urisReferencedByEntrypoints.has(file.uri),
```
lydell, line 257: - Files in `tests/` are tests for sure.
- Among other files, find those with `main =`, or are in `"exposed-modules"`. They are entrypoints for sure. All modules referenced by entrypoints cannot use "test-dependencies".
- Any remaining files are explained by this comment.

src/providers/diagnostics/elmMakeDiagnostics.ts
```diff
@@ -192,63 +188,171 @@ export class ElmMakeDiagnostics {
...
+    const elmTestNotFound =
+      "'elm-test' (or 'elm-test-rs') is not available. Install via for example 'npm install -g elm-test'.";
+
+    // - If all entrypoints are covered by tests, we only need to run `elm-test make`.
+    // - Otherwise, call `elm make` for all entrypoints (if any).
+    // - Call `elm-test make` for all tests (if any), plus potential tests.
+    // - If there’s no `tests/` folder but files that _could_ be tests, try to
+    //   call `elm-test make` but fall back to `elm make` in case they’re not
+    //   tests and the user hasn’t got elm-test installed.
+    const results = await Promise.allSettled([
```
lydell, line 287: This is the compilation strategy.

Note that with this approach we try to call `elm make` with just the entrypoints again, which is _slightly_ faster than calling it with _all_ files.

src/providers/diagnostics/elmMakeDiagnostics.ts
```diff
@@ -192,63 +188,171 @@ export class ElmMakeDiagnostics {
...
+    ]);
+
+    const lines: IElmIssue[] = [];
+    const linesSet = new Set<string>();
+
+    for (const result of results) {
+      if (result.status === "fulfilled") {
+        continue;
+      }
+      const error = result.reason as unknown;
```
lydell, line 351: This collects errors both from the `elm make` run and the `elm-test make` run, and tries to avoid duplicate errors. If there’s a compile error in the source code, test for that code will trigger the same compile errors.

razzeee:
Ignore the CI, coveralls has been down for the last day(s?)

lydell:
@razzeee After some brainstorming on Slack we came up with a good algorithm. Implementing it felt good!

- For packages, "test-dependencies" is allowed everywhere but in "exposed-modules", which should be a perfect solution.
- For applications, we do our best:
  - Files with `main =` are entrypoints for sure. Any module referenced from an entry point cannot import "test-dependencies" (perfect).
  - Files in `tests/` are tests for sure and can import "test-dependencies" (perfect).
  - Remaining files _could_ be tests so they can import "test-dependencies" (tradeoff). This means that all Elm Make diagnostics should be correct (fully sound), but some errors for "test-dependencies" imports could be missing (not complete). That’s much better than <em>in</em>correct Elm Make diagnostics (not sound) but no missing (read: too many) "test-dependencies" imports errors ((over-)complete).
  
Bonus: We now support elm-test-rs, since it’s faster than elm-test (no Node.js 100-200 ms startup cost).

lydell:
I’ve tested this PR on the ~160k lines of Elm code at work.

- It seems to work just fine!
- We don’t have any tests inside `src/`, but we still get some files that are marked as “possibly tests” because we have some unused modules (which are ignored by elm-review).
- I measured where time is spent in `ElmMakeDiagnostics.checkForErrors`. <10 ms is spent by the function itself, all the rest is just executing `elm make` and `elm-test make`/`elm-test-rs make`.
- With elm-test-rs installed, running `elm make` plus `elm-test-rs make` in parallel takes ~150 ms.
- With elm-test installed instead of elm-test-rs, `elm make` plus `elm-test-rs make` takes ~400 ms (because of Node.js startup time). So using elm-test-rs is a big win!

@razzeee If you think this approach looks good, let me know and I’ll go ahead and remove all elm-tooling.json/entrypoints related stuff. Or tell me if you’d rather have that in a separate PR.

mpizenberg:
Slightly related since I read elm-test-rs usage could be interesting. It does not support the `--report=json` option for the make subcommand. Is that something that would be needed?

lydell:
@mpizenberg Thanks for letting me know about that! It appears my testing with elm-test-rs was way too quick.

Yes, `--report json` is needed. This PR is broken as-is.

razzeee:
Let's edit the readme and recommend elm-test-rs instead of elm-test? Do you see any problems with that?

Do we need to reword the setting for tests? Is it clear, that it can take any of the compatible test runner? Should we have two settings and pick what's available?

src/compiler/utils/elmUtils.ts
```diff
@@ -42,7 +40,7 @@ export function execCmdSync(
     });
   } catch (error) {
     connection.console.warn(JSON.stringify(error));
-    if (error.code && error.code === "ENOENT") {
+    if (error.code === "ENOENT") {
```
razzeee, line 43: How are you sure this will always exist? `Error` only has stack, message and name for me. Did you find another type that matches?

If so please let me know, as that would be relevant for https://github.com/elm-tooling/elm-language-server/pull/638

src/compiler/utils/elmUtils.ts
```diff
@@ -55,6 +53,56 @@ export function execCmdSync(
...
+        options,
+        cwd,
+        connection,
+        input,
+      );
+    }
+    connection.console.warn(JSON.stringify(error));
+    if (notFound) {
+      connection.window.showErrorMessage(
+        options.notFoundText + ` I'm looking for commands at '${cwd}'`,
```
razzeee, line 97: The sync version also prints the `cmd` that wasn't found.

src/compiler/utils/elmUtils.ts
```diff
@@ -55,6 +53,56 @@ export function execCmdSync(
...
 
+/** Options for execCmd */
+export interface IExecCmdOptions {
+  /** Text to add when command is not found (maybe helping how to install)
+   * Unlike the sync version, it’s required here (since `cmdStatic` has fallbacks).
+   */
+  notFoundText: string;
+}
+
+export async function execCmd(
```
razzeee, line 64: I think it might be worth exploring to get rid of the sync version and just have the async one, as it trims down on maintenance. But it should probably be done outside of this PR.

src/providers/diagnostics/elmMakeDiagnostics.ts
```diff
@@ -192,63 +188,171 @@ export class ElmMakeDiagnostics {
...
+      treeMap,
+      entrypointsForSure,
+    );
+
+    const urisReferencedByTestsForSure = this.getUrisReferencedByEntrypoints(
+      treeMap,
+      testFilesForSure,
+    );
+
+    const allEntrypointsCoveredByTestsForSure = entrypointsForSure.every(
```
razzeee, line 238: Can we find a better name here? Only saw that it's a bool due to debugging and not sure what it's for. (but I haven't read on, so this might be premature)

src/providers/diagnostics/elmMakeDiagnostics.ts
```diff
@@ -192,63 +188,171 @@ export class ElmMakeDiagnostics {
...
+
+    const forestFiles: Array<ISourceFile> = Array.from(treeMap.values());
+
+    const allFiles = forestFiles.some((file) => file.uri === sourceFile.uri)
+      ? forestFiles
+      : forestFiles.concat(sourceFile);
+
+    const projectFiles = allFiles.filter((file) => !file.isDependency);
+
+    const testFilesForSure = projectFiles.filter((file) => file.isTestFile);
```
razzeee, line 213: In a perfect implementation, all this detection should end up setting the `isTestFile` bools. But it's a great start anyway.

src/compiler/utils/elmUtils.ts
razzeee: Realized your handling it. I was probably already thinking of merging sync and async, where displaying the cmd dynamically would make more sense.

src/providers/diagnostics/elmMakeDiagnostics.ts
```diff
@@ -192,63 +188,171 @@ export class ElmMakeDiagnostics {
...
-        testOrMakeCommand,
-        testOrMakeCommandWithOmittedSettings,
-        options,
-        workspaceRootPath,
-        this.connection,
-      );
-      return [];
-    } catch (error) {
+    const elmNotFound =
+      "The 'elm' compiler is not available. Install via for example 'npm install -g elm'.";
```
razzeee, line 277: I think the wording `via for example` might be a bit odd?

src/providers/diagnostics/elmMakeDiagnostics.ts
```diff
@@ -192,63 +188,171 @@ export class ElmMakeDiagnostics {
...
+    const elmTestNotFound =
+      "'elm-test' (or 'elm-test-rs') is not available. Install via for example 'npm install -g elm-test'.";
```
razzeee, line 279: Same here

src/providers/diagnostics/elmMakeDiagnostics.ts
razzeee: I think it could be something like `onlyRunElmTest`.

src/compiler/utils/elmUtils.ts
lydell: All Node.js core errors have `.code`: https://nodejs.org/api/errors.html#errors_error_code_1 and https://nodejs.org/api/errors.html#errors_node_js_error_codes

But even if we caught some other error here, it doesn’t matter if `.code` is `undefined`. `undefined === "ENOENT"` returns `false`, so there’s no need for checking if the `.code` is truthy first.

src/providers/diagnostics/elmMakeDiagnostics.ts
lydell: Yeah it might be. English is not my native language. My thinking was:

- We only have space to show one command?
- Suggesting to install `elm-test` is not the _only_ thing you can do – you can also install elm-test-rs.
- Maybe we should suggest installing elm-test-rs, but there’s no simple command to do it.
- There are more ways of installing Elm than via `npm`: The installer, `brew`, `elm-tooling`.

mpizenberg: Something along the line of "Follow instructions at ... for installation"?

lydell:
> Let's edit the readme and recommend elm-test-rs instead of elm-test? Do you see any problems with that?

I think that would make sense! We could mention that it’s for better performance, and then link to https://github.com/mpizenberg/elm-test-rs/#no-automatic-module-description so that people are aware of the difference in `describe` strings and that they might need to update them. What do you say, @mpizenberg?

> Do we need to reword the setting for tests? Is it clear, that it can take any of the compatible test runner? Should we have two settings and pick what's available?

I don’t think two settings are needed. Either the user uses elm-test-rs or node-test-runner, but not both. If there were two settings we would need a third setting to choose between the two as well?

I think we can tweak the text to clarify that it’s the path to a test runner, not to elm-test/node-test-runner specifically. Maybe all we need is “The path to your elm-test <ins>or elm-test-rs</ins> exectuable.”

mpizenberg:
> the difference in describe strings and that they might need to update them

I was hoping to get https://github.com/elm-explorations/test/pull/163 merged soon, after the CI is updated https://github.com/elm-explorations/test/pull/174. That would tremendously help people who might need to change multiple duplicated names instead of failing at the first one.

Another possibility is to have compatibility for both in this release but let elm-test be the default, but with a note about elm-test-rs. Then later if there is good feedback from early adopters, and not too many bugs push for it as default if the perf are worth it.

src/providers/diagnostics/elmMakeDiagnostics.ts
razzeee: Agreed, I would actually recommend elm-tooling. Maybe we should start our wiki and link to that.

We could also explore downloading elm-tooling for you and using it to install deps automatically (as the c# implementation does it) or we could now build platform specific binaries, as that's a new feature the vscode marketplace now supports. But that has other downsides.

razzeee:
I actually want elm-test-rs as default. We likely need to do a RC anyway, just to be safe.

src/compiler/utils/elmUtils.ts
razzeee: I guess typescript only types it as https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error as it doesn't know we're in node.

lydell: Yep, that’s super annoying. In my own typescript code I often do this:

```ts
} catch (errorAny) {
	const error = errorAny as Error & { code?: string };
	if (error.code === "...") {
		// ...
	}
}
```

(Not 100% safe, but I don’t want to treat every single `try-catch` as `unknown`…)

mpizenberg:
You shall find a working binary (I fixed `elm-test-rs make --report json`) in the CI of the latest commit on master if you want to test with elm-test-rs locally.

src/providers/diagnostics/elmMakeDiagnostics.ts
mpizenberg: A markdown document inside a docs/ folders should be enough I think.

I feel like trying to install automatically will bring more trouble than it's worth.

src/providers/diagnostics/elmMakeDiagnostics.ts
```diff
@@ -344,4 +458,31 @@ export class ElmMakeDiagnostics {
       );
     }
   }
+
+  private getUrisReferencedByEntrypoints(
```
razzeee, line 462: This only ever walks the first level of the dependency graph. I would have expected to actually walk the whole graph.
Isn't that necessary?

src/providers/diagnostics/elmMakeDiagnostics.ts
```diff
@@ -344,4 +458,31 @@ export class ElmMakeDiagnostics {
...
+    treeMap: Map<string, ISourceFile>,
+    entrypoints: ISourceFile[],
+  ): Set<string> {
+    const stack: ISourceFile[] = entrypoints.slice();
+    const result = new Set<string>(entrypoints.map((file) => file.uri));
+
+    for (let i = 0; i < stack.length; i++) {
```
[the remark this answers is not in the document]
razzeee: Nevermind, just realized what's happening here.

src/providers/diagnostics/elmMakeDiagnostics.ts
razzeee: NVM

src/compiler/utils/elmUtils.ts
razzeee: I think I actually would like to do it as unknown, at least until TS can type it correctly.
But might be shortsighted of me.

lydell:
@razzeee elm-test-rs is super fast and really promising to start using in the language server, but it seems like it needs a few tweaks before being usable. This is a blocker: https://github.com/mpizenberg/elm-test-rs/issues/103

Should we remove elm-test-rs support from this PR for now and revisit it later? Or should we keep this open for as long as it takes for us to fix things in elm-test-rs?

mpizenberg:
@lydell there is also the problematic fact we discussed with @razzeee that the VSCode extension cannot match the tests with the labels in the json output of elm-test-rs (missing the module name top-level label). I want to spend some time on this, but it won't happen soon (not in the next 2 weeks), so I'd prefer if the integration could be delayed.

razzeee:
Let's move on without etrs for now. Was nice to include, but should probably be different merges anyway.

That being said, the other issue is not connected to this. Both are separate code paths.

src/compiler/utils/elmUtils.ts
lydell: That should be a separate PR then. I can’t find any try-catch in the code base that actually treats the caught errors as `unknown`. Btw, there is an option these days to enforce them as `unknown`: https://www.typescriptlang.org/tsconfig#useUnknownInCatchVariables

razzeee: You are correct. Have a look at https://github.com/elm-tooling/elm-language-server/pull/638/files ;)

src/providers/diagnostics/elmMakeDiagnostics.ts
lydell: I reverted to the same message as in `main`.

src/providers/diagnostics/elmMakeDiagnostics.ts
lydell: I reverted to the same message as in `main`.

lydell:
Companion PR for elm-language-client-vscode: https://github.com/elm-tooling/elm-language-client-vscode/pull/215

src/compiler/utils/elmUtils.ts
lydell: Interesting blog post about it: https://dev.to/jdbar/the-problem-with-handling-node-js-errors-in-typescript-and-the-workaround-m64

lydell: I think this approach ended up quite nice: https://github.com/elm-tooling/elm-tooling-cli/pull/75/files

razzeee:
Thank you for working on this.

PR #641, opened 2021-09-24
Add function argument from missing declaration

### LINKED ISSUE #636: Introduce new argument to function from missing declaration
When a declaration is missing, it should be possible to introduce that as an argument to the enclosing function.

## Expected Behavior
When requesting an action, or fix, on a `Diagnostics.MissingValue`, there should be an option in the menu to introduce a new argument to the function.
It should also add the inferred type signature if the function already has a type annotation.

## Current Behavior
Requesting the fix only provides an option to `create local function`.

## Possible Solution

I have a work-in-progress implementation of a possible solution:
Screen Recording 2021-09-16 at 4 03 06 PM

## Context
This is really useful when you for example need to thread a value from the model through many functions and need to update arguments frequently.

## Other comments

I was surprised that `create local function` does not create a declaration within a let-block, but rather a new top level function. I can see there is an action for "external functions", but i'm not sure when this is triggered.
Would like to hear if i'm missing something here, or if there is some rework that should be done around these actions.

Would also like some input on some UX details like the name of the action in the menu.

Gauteab:
closes #636 
Screen Recording 2021-09-24 at 7 56 32 PM (1)


razzeee:
Awesome, can you do some basic tests? Would you need any help with that?

src/providers/codeAction/index.ts
```diff
@@ -10,3 +10,4 @@ import "./addMissingRecordFieldCodeAction";
 import "./addMissingCaseBranchesCodeAction";
 import "./elmReviewFixAction";
 import "./addMissingUnionType";
+import "./addNewFunctionParamterer";
```
razzeee, line 13: Typo in `Parameter`

src/providers/codeAction/addNewFunctionParamterer.ts
```diff
@@ -0,0 +1,104 @@
...
+  ) {
+    return TreeUtils.getAllAncestorsOfType("value_declaration", nodeAtPosition)
+      .map((valueDeclaration) =>
+        getActionsForValueDeclaration(valueDeclaration, nodeAtPosition, params),
+      )
+      .filter((e): e is ICodeAction => e != undefined);
+  }
+}
+
+function getActionsForValueDeclaration(
```
razzeee, line 43: Should probably be private

src/providers/codeAction/addNewFunctionParamterer.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  const functionName = valueDeclaration.firstChild?.firstChild?.text;
+  return CodeActionProvider.getCodeAction(
+    params,
+    `Add new parameter to "${functionName}"`,
+    edits,
+  );
+}
+
+function getEditsForSignatureUpdate(
```
razzeee, line 75: Should probably be private

src/providers/codeAction/addNewFunctionParamterer.ts
```diff
@@ -0,0 +1,104 @@
...
+
+const errorCodes = [Diagnostics.MissingValue.code];
+const fixId = "add_new_function_parameter";
+
+CodeActionProvider.registerCodeAction({
+  errorCodes,
+  fixId,
+  getCodeActions: (params: ICodeActionParams) =>
+    getActions(params, params.range),
+  getFixAllCodeAction: (_: ICodeActionParams): ICodeAction | undefined =>
```
razzeee, line 16: Just do `getFixAllCodeAction: (): ICodeAction | undefined => undefined,` here

src/providers/codeAction/addNewFunctionParamterer.ts
```diff
@@ -0,0 +1,104 @@
...
+    nodeAtPosition,
+    valueDeclaration,
+  );
+
+  edits.push(
+    TextEdit.insert(valueParameterPosition, nodeAtPosition.text + " "),
+  );
+
+  const functionName = valueDeclaration.firstChild?.firstChild?.text;
+  return CodeActionProvider.getCodeAction(
```
razzeee, line 68: I think we need to guard this from the potentially undefined `functionName`

src/providers/codeAction/addNewFunctionParamterer.ts
```diff
@@ -0,0 +1,104 @@
...
+      .filter((e): e is ICodeAction => e != undefined);
+  }
+}
+
+function getActionsForValueDeclaration(
+  valueDeclaration: SyntaxNode,
+  nodeAtPosition: SyntaxNode,
+  params: ICodeActionParams,
+): CodeAction | undefined {
+  const lastPattern = valueDeclaration?.firstChild?.lastChild;
```
razzeee, line 48: Can we use a name like `lastFunctionParameter` for this?

razzeee:
Everything else LGTM

Gauteab:
Thanks for the feedback. I'll try to find the time to finish up and write some tests soon :)

razzeee:
I've fixed most of the tests, I'm not sure what the last one is supposed to be testing.

Gauteab:
Thanks! There was a mistake in the expected source of the last test. It checks that you can add a missing declaration that appears in the body of a local/inner function to the parameters of the enclosing outer function.

razzeee:
Should be good to go then?

Gauteab:
Theres one more thing that is not totally correct.

When you use `TypeChecker.findType` and the input can be "any" type, it seems that it returns a `TVar` with the text `a`, regardless of the context.

```elm
f : a -> a
f a =
    always a x
```
In the above example, the inferred type of `x` is `a`, but it should be something more general, like `b`.

When the missing declaration can be any type, I think it makes most sense to use the variable name as its type name, so the result of adding `x` to the signature should be:
```elm
f : a -> x -> a
f a x =
    always a x
```

Up until now I've been using the variable name whenever `findType` returned a `TVar`, but that is not correct when the type of the missing declaration is that of an already bound type variable.
Theres also another edge case with using the variable name as the type name when the TVar is a constrained variable, but thats simple enough to fix.

Is this the correct behavior for `findType`?
Is there a simple way to know if the expression I'm checking can be any type?
I feel like it would be a hack to go through the type signature and see if there is an `a` there already 🤔

razzeee:
@jmbockhorst might also have an oppinion on this. I'll try to dig into the code in the next days.

razzeee:
Do you think this is big enough of a problem to keep this from being merged? I would like to do a RC version.

jmbockhorst (approved):
This looks good to me. @Gauteab I'm not sure what you are talking about with the behavior of `findType`. It seems to be working as I would expect and as you said.

Issue #635, opened 2021-09-16
elm-language-server won't work with vim/ALE


jpalardy:
Hi there!

Thanks for elm-language-server! It works great with vscode 😄 

On the other hand, it seems broken with vim + ALE… here's what I found:

## Research

ALE doesn't send the `range` property … it always send the whole file. (not the best, but simple)

![Screen Shot 2021-09-15 at 23 35 01](https://user-images.githubusercontent.com/18142/133561956-d953f387-306d-4941-95d6-715529378f51.png)

Meanwhile, LSP makes some assumptions about the message fields:

![Screen Shot 2021-09-15 at 23 36 42](https://user-images.githubusercontent.com/18142/133562253-870b6fba-cf6c-4d53-b68b-a2796a1befc1.png)

(although, I didn't follow all the code here, I'll defer to you)

The [spec](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/) seems to allow for optional `range` (possibly for backward compatibility?)

![Screen Shot 2021-09-15 at 23 40 57](https://user-images.githubusercontent.com/18142/133562903-46630628-c23c-47a4-b2ac-9cc6e56db3c5.png)

## Steps to Reproduce (for bugs)

1. open elm file with vim
2. make a change
3. see everything become a "parsing error"

## Your Environment

* Version used: 2.2.1
* Editor used: vim 8.2
* Environment name and version (e.g. node.js 5.4): node v16.9.1, npm 7.21.1
* Operating System and version: macos (and linux in vagrant)

jpalardy:
If you need any more information, or want me to try something — I'll be available.

Let me know — and thanks again!

jpalardy:
Saw #503, which sent me down the rabbit hole 😅

razzeee:
Interesting, I actually thought the server has the last word via https://github.com/elm-tooling/elm-language-server/blob/main/src/capabilityCalculator.ts#L45

https://microsoft.github.io/language-server-protocol/specification#textDocument_synchronization_sc

jpalardy:
Ah… might be ^

2 thoughts:
- isn't accepting "the whole file content" easier? (I'm willing to pitch it, if that helps)
- what if ALE "cheated" and sent a `range` that covered the whole file — would that work? (I might need some help on what exact values to pass).  If we can make this work, I'll push the patch on ALE on my side (no changes for this repo)

razzeee:
[...]

PR #642, opened 2021-09-26
Fix vim-ale when `range` missing from `contentChanges`

jpalardy:
As discussed in #635

If `contentChanges` doesn't contain a `range` property, we assume the whole file content was sent.

This follows from the details of [DidChangeTextDocument](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#textDocument_didChange), specifically:

133562903-46630628-c23c-47a4-b2ac-9cc6e56db3c5

It's assumed the text goes from `{line: 0, character: 0}` to `{line: textSplitInLines.length, character: 0}`. This seems to follow the [details about range](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#range) from the spec.

### DIFF
--- a/src/providers/astProvider.ts
+++ b/src/providers/astProvider.ts
@@ -81,6 +81,17 @@ export class ASTProvider {
       for (const change of params.contentChanges) {
         if ("range" in change) {
           tree?.edit(this.getEditFromChange(change, tree.rootNode.text));
+        } else {
+          const currentText = tree?.rootNode.text;
+          if (currentText) {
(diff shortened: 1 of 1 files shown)


razzeee:
I had a debugging session and it looks good. All things I was considering to improve don't seem worth the hassle. So good work. I'll merge this now.

jpalardy:
Fantastic ^

Ping me on any problems if it comes up 👍

razzeee:
Will do, thank you

PR #656, opened 2021-11-13
Ignore suppressed errors in elm-review's output

### LINKED ISSUE #655: elm-review suppress support
I am adding a new feature to `elm-review` which allows users to suppress errors, available in a beta version (`npm install elm-review@beta`), which impacts the JSON output. I made a short description for it [here](https://github.com/jfmengels/elm-review#temporarily-suppressing-errors), but intend to announce it (still as a beta) in the coming days.

## Expected Behavior

I have added two fields on the object for each error, `suppressed` and `originallySuppressed`.

I have described them in the [tooling integration document](https://github.com/jfmengels/node-elm-review/blob/master/documentation/tooling-integration.md#review-errors), but in short the editor should not report the errors where `suppressed` is falsy (and nothing more as far as I can tell).

The reason why I have chosen this format is so that it is easy to include or exclude these errors (without forcing tools to sort them, etc), should they want to do something with them. For instance, there could be a setting/checkbox/shortcut to have the suppressed errors shown as warnings (or a different color if available), should users want that (to tackle the suppressed errors for instance).

## Current Behavior

Currently, the LSP doesn't filter the suppressed errors, meaning suppressed errors will be reported to the user.

## Steps to Reproduce

1. Clone https://github.com/jfmengels/elm-spa-example (note that it's my fork, not the original repo)
2. Checkout the `suppressed-errors` branch
3. Open the project with the IDE
4. Configure the IDE to show `elm-review` errors
5. You'll have `elm-review` errors reported, but once fixed, you should see none.

Since the feature is behind a beta version, I don't think there is a need to release a new version quickly. I'll try to submit a PR very soon. If you think this approach is disturbing (it is a bit of a breaking change...?), I apologize, and please let me know if you have better alternatives in mind.

jfmengels:
Fixes #655: Ignore suppressed errors from `elm-review`'s JSON output.

This should be all that's necessary, but I don't know how to try out this patch locally :thinking:

### DIFF
--- a/src/providers/diagnostics/elmReviewDiagnostics.ts
+++ b/src/providers/diagnostics/elmReviewDiagnostics.ts
@@ -56,6 +56,7 @@ interface IError {
     range: IRegion;
     string: string;
   }[];
+  suppressed?: boolean;
 }
 
 interface IRegion {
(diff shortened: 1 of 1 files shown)


src/providers/diagnostics/elmReviewDiagnostics.ts
```diff
@@ -172,26 +173,28 @@ export class ElmReviewDiagnostics {
...
-                  code: "elm_review",
-                  fixes: (error.fix || []).map((fix) => ({
-                    string: fix.string,
-                    range: toLsRange(fix.range),
-                  })),
-                },
-              })),
+              errors
+                .filter((error: IError) => !error.suppressed)
+                .map((error: IError) => ({
```
jfmengels, line 178: I reindented the code, but the only addition in this block is the addition of the the filter function.

razzeee:
Thank you

jfmengels:
Alright, good to know :smile: 
Thank you :pray:

Issue #687, opened 2022-03-23
`elm make` Fails After Starting Elm Language Server


tmciver-mackey:
Emacs' integration with `elm-language-server` does not work as server responds with an error for all interactions.

## Expected Behavior

No error reported by `elm-language-server`.

## Current Behavior

Starting lsp in a buffer containing an Elm source file with elm-mode activated causes the following text to be displayed in *lsp-log*:
```
Command "/home/tim/.nvm/versions/node/v16.4.0/bin/elm-language-server --stdio" is present on the path.
Command "/home/tim/.nvm/versions/node/v16.4.0/bin/elm-language-server --stdio" is present on the path.
Found the following clients for /home/tim/workspace/front-end/elm/src/Layout/SearchBox.elm: (server-id elm-ls, priority -1)
The following clients were selected based on priority: (server-id elm-ls, priority -1)
Loading Elm tree-sitter syntax from ../../.nvm/versions/node/v16.4.0/lib/node_modules/@elm-tooling/elm-language-server/out/tree-sitter-elm.wasm
Found 2 elm.json files for workspace /home/tim/workspace/front-end
Found 1 unique elmWorkspaces for workspace /home/tim/workspace/front-end
Starting language server for folder: file:///home/tim/workspace/front-end
Elm version 0.19.1 detected.
Reading elm.json from /home/tim/workspace/front-end/elm.json
{"shortMessage":"Command failed with exit code 1: elm make","command":"elm make","escapedCommand":"elm make","exitCode":1,"stdout":"","stderr":"-- NO INPUT --------------------------------------------------------------------\n\nWhat should I make though? I need specific files like:\n\n    elm make src/Main.elm\n    elm make src/This.elm src/That.elm\n\nI recommend reading through https://guide.elm-lang.org for guidance on what to\nactually put in those files!\n\n","failed":true,"timedOut":false,"isCanceled":false,"killed":false}
Error parsing files for /home/tim/workspace/front-end/elm.json:
Error: ENOENT: no such file or directory, open '/home/tim/.elm/0.19.1/packages/ContaSystemer/elm-menu/1.1.1/elm.json'
Notification handler 'textDocument/didOpen' failed with message: Cannot read property 'getTree' of undefined
TypeError: Cannot read property 'getTree' of undefined delayed processing of request
```

## Steps to Reproduce (for bugs)

1. Install `elm-laguage-server` globally with:
        `npm i -g @elm-tooling/elm-language-server`
[...]

Issue #646, opened 2021-10-09
Language server fails to start if dependencies are not in ~/.elm folder


skovmand:
First of all thank you for your work on this project!

I have moved to a new machine since I last worked on my elm project. Therefore the packages for my project were not available in the `~/.elm` folder. This made the LS crash on start with a confusing error.

## Expected Behavior
The language server could give a better error message, providing a possible solution.

## Current Behavior
I get the output in the VSCode console:

```
[Info  - 11:25:04 AM] Loading Elm tree-sitter syntax from ../../../../../.vscode/extensions/elmtooling.elm-ls-vscode-2.2.1/server/out/tree-sitter-elm.wasm
[Info  - 11:25:04 AM] Found 1 elm.json files for workspace /Users/<user>/code/elm-project
[Info  - 11:25:04 AM] Found 1 unique elmWorkspaces for workspace /Users/<user>/code/elm-project
[Info  - 11:25:04 AM] Starting language server for folder: file:///Users/<user>/code/elm-project
[Info  - 11:25:04 AM] Elm version 0.19.1 detected.
[Info  - 11:25:04 AM] Reading elm.json from /Users/<user>/code/elm-project/elm.json
[Warn  - 11:25:04 AM] {"shortMessage":"Command failed with exit code 1: elm make","command":"elm make","escapedCommand":"elm make","exitCode":1,"stdout":"","stderr":"-- NO INPUT --------------------------------------------------------------------\n\nWhat should I make though? I need specific files like:\n\n    elm make src/Main.elm\n    elm make src/This.elm src/That.elm\n\nI recommend reading through https://guide.elm-lang.org for guidance on what to\nactually put in those files!\n\n","failed":true,"timedOut":false,"isCanceled":false,"killed":false}
[Error - 11:25:04 AM] Error parsing files for /Users/<user>/code/elm-project/elm.json:
Error: ENOENT: no such file or directory, open '/Users/<user>/.elm/0.19.1/packages/elm/browser/1.0.2/elm.json'
[Error - 11:25:04 AM] Notification handler 'textDocument/didOpen' failed with message: Cannot read property 'getTree' of undefined
[Error - 11:25:04 AM] Request textDocument/codeAction failed.
  Message: Request textDocument/codeAction failed with message: Cannot read property 'getTree' of undefined
  Code: -32603 
[Error - 11:25:04 AM] Request textDocument/foldingRange failed.
[...]

Issue #640, opened 2021-09-21
elm-language-server stopped working


3nt3:
## Expected Behavior
* it should just work (e.g. auto complete in vs code)

## Current Behavior
* nothing happens when starting to type - only general word suggestions
* there are hundreds of errors like this in the console: ![error](https://i.imgur.com/tcJx9WY.png)
## Your Environment

* Version used: 2.2.1
* Editor used: code-oss, version 1.58.0 (commit `2d23c42a936db1c7b3b06f918cde29561cc47cd6`)
* Environment name and version (e.g. node.js 5.4): node.js v16.9.1
* Operating System and version: 5.14.6-arch1-1

3nt3:
yep everything is installed

razzeee:
Can you please post a complete log? Or at least to the point where the error starts?

Relevant: 
https://github.com/elm-tooling/elm-language-server/issues/518
https://github.com/elm-tooling/elm-language-server/issues/624

3nt3:
here's a log http://ix.io/3zGV

razzeee:
There should be another log in the output view. You'll need to use the dropdown to select `Elm (name of you project)`. That should hopefully have more context.

razzeee:
Can you get back with the logfile?

3nt3:
I'm sorry, can you maybe send me a screenshot of where that button should be located?

razzeee:
1. F1 > Output: Focus on Output View
2. Choose dropdown on the right side

![image](https://user-images.githubusercontent.com/5943908/147499180-50f75921-62b3-4a6c-849c-7d410153079e.png)

razzeee:
were you able to solve this?

Issue #518, opened 2021-02-12
utils.getElmVersion returns empty version


jlocans:
## Expected Behavior
~~Extension should work (look for Elm packages in right directory)~~
utils.getElmVersion should return Elm version that is installed. `0.19.1` in this case.

## Current Behavior
UPDATED: Found that `utils.getElmVersion` is returning empty value instead of version.

-------------------------

Opening the elm-spa-example project. The only thing that seems to work is syntax highlighting.
One of the first errors in output is 
`Error: ENOENT: no such file or directory, open 'C:\Users\janis\AppData\Roaming\elm\packages\NoRedInk\elm-json-decode-pipeline\1.0.0\elm.json'`
The file is actually located in
`C:\Users\janis\AppData\Roaming\elm\0.19.1\packages\NoRedInk\elm-json-decode-pipeline\1.0.0\elm.json`
See the `0.19.1` version part.

The weird thing is that it worked the first time I installed it (moments before I reinstalled it). Basically I installed the extension. It worked. Then noticed numbers next to elm files in file explorer and was wondering if that is extension feature or vscode feature (which I had just updated as well) as I couldn't understand what they meant. Fastest way to check from where the features comes? Uninstall extension. Uninstalled it, numbers were gone. Ok, so that is extensions feature. Installed it back and it doesn't work any more.

Then I found the `ENOENT` error and that I had two version folders inside `AppData\Roaming\elm\packages\` - `0.19.0` and `0.19.1`.
Uninstalled both Elm versions.
Deleted elm-spa-example project
Deleted elm folder in `AppData\Roaming`
Installed Elm `0.19.1`
Cloned project again and built it, but the path is still as `AppData\Roaming\elm\0.19.1\packages`

## Possible Solution
No solution, but seems that extensions is looking for packages in the wrong directory.

## Your Environment
* Version used: v2.0.3

vscode info:
Version: 1.53.1 (system setup)
Commit: 5d424b828ada08e1eb9f95d6cb41120234ef57c7
Date: 2021-02-08T23:29:42.785Z
Electron: 11.2.1
Chrome: 87.0.4280.141
Node.js: 12.18.3 EDIT: Updated to v14.15.5
V8: 8.7.220.31-electron.0
OS: Windows_NT x64 10.0.17134

## Full log
```log
[...]

PR #690, opened 2022-04-23
Improve error handling for broken package cache

klassiker:
The current error handling for the workspace initialization silently fails, providing no visible error indication. This leaves extensions in an unusable state, just reporting `TypeError: Cannot read property 'getTree' of undefined` in the console because `this.forest` was never assigned https://github.com/elm-tooling/elm-language-server/blob/b22398b3398b20640cbdd9cc6d6b70a14826b800/src/compiler/program.ts#L388-L390

The main cause for this failure is a broken package cache, most likely caused by incompatible or multiple elm installations.

Similar things happened to multiple users, #518 #646 #687 and elm-tooling/elm-language-client-vscode#233. #640 is also most likely caused by this.

The final warning might need adjustments, as the LS is used in multipled IDEs.

The information message for the package cache is already very long, but users should also verify that the mentioned path actually does not exist and that there is no other location inside the cache providing that package.

### DIFF
--- a/src/compiler/program.ts
+++ b/src/compiler/program.ts
@@ -395,9 +395,9 @@ export class Program implements IProgram {
         }
 
         this.connection.window.showInformationMessage(
-          "The package cache is probably broken. Try a restart after removing '~/.elm' or '%APPDATA%\elm'."
-          + "If the error still occurs, try running 'elm init' in a different folder."
-          + "If the error appears again, check your PATH for multiple elm installations and verify your installed version",
+          "The package cache is probably broken. Try a restart after removing '~/.elm' or '%APPDATA%\\elm'." +
+            "If the error still occurs, try running 'elm init' in a different folder." +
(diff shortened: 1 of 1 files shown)


src/compiler/program.ts
```diff
@@ -427,6 +444,13 @@ export class Program implements IProgram {
         );
       }
     }
+
+    if (this.forest === null) {
+      this.connection.window.showWarningMessage(
+        `Extension will not work at all: workspace initialization failed for ${pathToElmJson}`
+        + "For more information, check the console output (F1 > Output, dropdown on the right, 'Elm (project name)')"
```
razzeee, line 451: I think one way would be to make clear, that this is an instruction for VSCode.

Or we could just say check your extension logs.

razzeee:
It seems like the test suite doesn't like this behavior, we probably don't initialize this correctly in some tests, where we're not testing this code path.

razzeee:
Thank you

PR #691, opened 2022-05-05
Fix for reference count of function in a let, with a type annotation.

jbuedel:
This fixes a bug where values in a `let...in` have a zero reference count, if they have a type annotation.

This gif demonstrates the problem. Watch how the count changes from 2 to 0 depending on the presence of the type annotation. 
Code_NklwL66TI0

### DIFF
--- a/src/compiler/typeChecker.ts
+++ b/src/compiler/typeChecker.ts
@@ -449,6 +449,21 @@ export function createTypeChecker(program: IProgram): TypeChecker {
           diagnostics: [],
         };
       }
+
+      // This is for values, with type annotation, but *inside* a let.
+      if (nodeParent.nextNamedSibling?.type === "value_declaration") {
+        // A value_declaration with a type annotation got it's type_annotation node
+        // added. So, we confirm the value_declaration does indeed follow, and return
+        // type node text with value declaration node.
+        return {
+          symbol: {
+            name: nodeText,
+            node: nodeParent.nextNamedSibling.children[0], // the function_declaration_left
+            type: "Function",
+          },
+          diagnostics: [],
+        };
+      }
     } else if (
       (nodeParentType === "exposed_type" &&
         nodeParent.parent?.parent?.type === "module_declaration") ||


src/compiler/typeChecker.ts
```diff
@@ -447,6 +447,23 @@ export function createTypeChecker(program: IProgram): TypeChecker {
           diagnostics: [],
         };
       }
+
+      // This is for values, with type annotation, but *inside* a let.
+      if (
+        nodeAtPosition.parent?.nextNamedSibling?.type === "value_declaration"
```
jmbockhorst, line 453: `nodeAtPosition.parent?` can be replaced with `nodeParent`, for a small perf improvement. Same for below.

src/compiler/typeChecker.ts
```diff
@@ -447,6 +447,23 @@ export function createTypeChecker(program: IProgram): TypeChecker {
...
+      ) {
+        // A value_declaration with a type annotation got it's type_annotation node
+        // added. So, we confirm the value_declaration does indeed follow, and return
+        // type node text with value declaration node.
+        return {
+          symbol: {
+            name: nodeAtPosition.text,
```
jmbockhorst, line 460: Replace `nodeAtPosition.text` with `nodeText`.

jmbockhorst (changes requested):
Just a few minor changes. Also, it would be great if you could write a simple test in `functionDefinition.test.ts`.

razzeee:
This needs formatting with prettier. Should this then still be a draft?

jbuedel:
> This needs formatting with prettier. Should this then still be a draft?

Hey @razzeee I don't understand. Are you suggesting I take it out of draft mode until the formatting is corrected? I put it in draft mode to indicate I was still working on it. (Which includes adding that unit test, formatting, and whatever else comes up.)

razzeee:
Keep it in draft, if you want to keep working on it. Sometimes people forget to change it :)

jbuedel:
@razzeee It seems that my unit test is not actually even calling the bug fix code that I added. Am I interpreting this correctly?

razzeee:
Yes, your code is not reached. It seems like it's due to the first `--X` changing the code/tree in some significant way.

razzeee:
We know this fixes the problem. I've just retested it.

I've removed the test now, as it's not proving that's the case, as the added comment below the type annotation seems to change the parsed tree and thus invalidates the test.

PR #725, opened 2022-07-16
Add exposing type with variants action

IzumiSy:
Currently, expose/unexpose action does not cover exposing types with variants, but I want that capability.

This PR implements it.


razzeee:
Thank you

Issue #686, opened 2022-03-21
Progress specification


Massolari:
There is [this plugin](https://github.com/j-hui/fidget.nvim) I'm trying to use in Neovim that displays when the language-server is indexing files.
It works with a variety of language servers, like tsserver, rust analyzer, null-ls, haskell, and some others, but it does not with elm-language-server.
In the README it says that the language server needs to implement the [progress specification](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#progress)

## Expected Behavior
When the language server is indexing files, it should appear in the editor, like:

![image](https://github.com/j-hui/fidget.nvim/raw/media/gifs/fidget-demo-rust-analyzer.gif?raw=true)

## Current Behavior
Nothing shows up

## Possible Solution
Check if we implement this specification and if we do, investigate why the plugin doesn't work

## Steps to Reproduce (for bugs)

1. Install the plugin
2. Open an Elm file

## Context
In big projects, elm-language-servers takes a while to index all the files, so, it'd be nice if we have some indication of the progress

## Your Environment

* Version used: 2.4.1
* Editor used: Neovim 0.6.1
* Environment name and version (e.g. node.js 5.4): Node v17.7.1
* Operating System and version: MacOS 12.3

IzumiSy:
For vim-lsp, I opened a PR that approaches this issue.

https://github.com/prabirshrestha/vim-lsp/pull/1335

razzeee:
Server side this has been implemented since the initial release I think. So this problem is likely client side.

PR #726, opened 2022-07-16
Uses createWorkdoneProgress for indexing notification

IzumiSy:
This probably solves #686

I also confirmed this working with [lightline-lsp-progress](https://github.com/micchy326/lightline-lsp-progress)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -57,10 +57,7 @@ const connection = container.resolve<Connection>("Connection");
 let server: ILanguageServer;
 
 connection.onInitialize(
-  async (
-    params: InitializeParams,
-    cancel,
-  ): Promise<InitializeResult> => {
+  async (params: InitializeParams, cancel): Promise<InitializeResult> => {
     await Parser.init();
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -57,11 +57,7 @@ const connection = container.resolve<Connection>("Connection");
 let server: ILanguageServer;
 
 connection.onInitialize(
-  async (
-    params: InitializeParams,
-    cancel,
-    progress,
-  ): Promise<InitializeResult> => {
+  async (params: InitializeParams, cancel): Promise<InitializeResult> => {
```
razzeee, line 60: I'm wondering if this still works when you start something, that has multiple progess bars, as it starts multiple elm servers, like multiple workspaces in one vscode project for e.g.

I think that was the original reason for handing it in here

IzumiSy: Now, I found this PR (https://github.com/elm-tooling/elm-language-server/pull/214), and it said like you told. Let me dig more of it why this does not work for me.

IzumiSy: I found this is caused by implementation of vim-lsp. Closing PR.

PR #727, opened 2022-07-17
Tweak progress message

IzumiSy:
See comments on diff

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -123,7 +123,7 @@ export class Server implements ILanguageServer {
   }
 
   public async init(): Promise<void> {
-    this.progress.begin("Indexing Elm", 0);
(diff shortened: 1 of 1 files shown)


src/server.ts
```diff
@@ -137,7 +137,7 @@ export class Server implements ILanguageServer {
             const avgIndexed =
               all.reduce((sum, { indexedPercent }) => sum + indexedPercent, 0) /
               all.length;
-            this.progress.report(avgIndexed, `${Math.round(avgIndexed)}%`);
+            this.progress.report(Math.round(avgIndexed), "Indexing");
```
IzumiSy, line 140: I think report message should be the detail of progress.

Currently, my environment using lightline-lsp-progress shows like `Indexing Elm 50% 50%`, but this looks redundunt.

Issue #728, opened 2022-07-17
ElmLS does not send `textDocument/publishDiagnostics` when entrypoint module is depended by tests and `elm-test` is installed locally.


IzumiSy:
~This only happens with vim-lsp environment. I cannot reproduce this issue on VSCode.~
I confirmed the same behavior on VSCode as well. See repro steps.

## Expected Behavior
ElmLS sends `textDocument/publishDiagnostics` to show an error

## Current Behavior
No diagnostic appears.

## Possible Solution
Possible reason of this issue is that `elm-test` cannot discover elm binary when it is installed locally, as vim-lsp logging says.
```
Mon Jul 18 02:21:21 2022:["<---",1,"elm-language-server",{"response":{"method":"window/logMessage","jsonrpc":"2.0","params":{"message":"{\"shortMessage\":\"Command failed with exit code 1: ./node_modules/.bin/elm-test make tests/Tests/Main.elm --report json\",\"command\":\"./node_modules/.bin/elm-test make tests/Tests/Main.elm --report json\",\"escapedCommand\":\"\\\"./node_modules/.bin/elm-test\\\" make \\\"tests/Tests/Main.elm\\\" --report json\",\"exitCode\":1,\"stdout\":\"\",\"stderr\":\"Cannot find elm executable, make sure it is installed.\\n(If elm is not on your path or is called something different the --compiler flag might help.)\\n\",\"failed\":true,\"timedOut\":false,\"isCanceled\":false,\"killed\":false}","type":2}}}]
Mon Jul 18 02:21:21 2022:["<---",1,"elm-language-server",{"response":{"method":"window/logMessage","jsonrpc":"2.0","params":{"message":"Received an invalid json, skipping error.","type":2}}}]
Mon Jul 18 02:21:21 2022:["<---",1,"elm-language-server",{"response":{"method":"window/logMessage","jsonrpc":"2.0","params":{"message":"Received an invalid json, skipping error.","type":2}}}]
Mon Jul 18 02:21:21 2022:["<---",1,"elm-language-server",{"response":{"method":"window/logMessage","jsonrpc":"2.0","params":{"message":"Received an invalid json, skipping error.","type":2}}}]
```
I have confirmed that using `--compiler` option when calling `elm-test` internally for the environment that has installed elm toolchains locally will fix this problem. 

My opened PR is https://github.com/elm-tooling/elm-language-server/pull/729

#### 0. Install elm, elm-test locally.
```sh
[...]

PR #729, opened 2022-07-17
Use --compiler option for elm-test when elmPath is specified

IzumiSy:
#728

### DIFF
--- a/src/providers/diagnostics/elmMakeDiagnostics.ts
+++ b/src/providers/diagnostics/elmMakeDiagnostics.ts
@@ -269,9 +269,8 @@ export class ElmMakeDiagnostics {
         "json",
       ];
 
-      if (settings.elmPath) {
-        args.push("--compiler");
-        args.push(settings.elmPath);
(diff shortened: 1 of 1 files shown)


src/providers/diagnostics/elmMakeDiagnostics.ts
```diff
@@ -261,12 +261,21 @@ export class ElmMakeDiagnostics {
...
-    ];
+    const argsElmTest = (files: Array<ISourceFile>): Array<string> => {
+      const args = [
+        "make",
+        ...files.map(fileToRelativePath),
+        "--report",
+        "json",
+      ];
+
+      if (settings.elmPath) {
```
razzeee, line 272: Might make sense to keep this similar to the other places, for e.g.:

```
    if (settings.elmPath.trim().length > 0) {
      cmdArguments.push("--compiler", settings.elmPath);
    }
```

Or the one in util.ts, but that might be unusable due to it also appending the files.

PR #816, opened 2022-11-01
Document elm-analyse integration configuration

anagrius:


README.md
```diff
@@ -103,6 +103,13 @@ If all you had in `elm-tooling.json` was `"entrypoints"`, you can safely remove
 
 Currently, no configuration at all is needed.
 
+### Configuring linting rules
+
+Linting is handled by [Elm-Analyse](https://stil4m.github.io/elm-analyse/).
```
razzeee, line 108: That's not true anymore, while we support the same config file, we've implemented these linters from scratch. I think we also omited some, that we thought we're unhelpful.

anagrius: It would probably be good to have those listed and documented.
Any reference to a file where I can find the linting rule implementations?

razzeee: Not the implementation, but the config interface https://github.com/elm-tooling/elm-language-server/blob/main/src/providers/diagnostics/elmAnalyseJsonService.ts#L5

anagrius: Thanks. I will make another proposal for documenting the rules.
Going through the list I see there is an entry for `MissingTypeAnnotation`:

https://github.com/elm-tooling/elm-language-server/blob/eb912794a79e4a786bca2f3ec8dcb0706fbeecaa/src/providers/diagnostics/elmAnalyseJsonService.ts#L36

Elm-Analyse has the `NoTopLevelSignature` check, which (seems) to be the same thing as `MissingTypeAnnotation` does. Maybe I am missing something.

anagrius: I have made another PR, and will close this one. The question about `MissingTypeAnnotation` vs `NoTopLevelSignature` stands though.

Please refer to:
https://github.com/elm-tooling/elm-language-server/pull/821

PR #819, opened 2022-11-04
Don't try to get code actions when there is a top level parsing error

jmbockhorst:
During a top level parsing error (like an unclosed if else statement), sometimes many hundreds of incorrect diagnostics would be created, due to invalid tree syntax state. The language server would get stuck trying to calculate "fix all" code actions, since it has to loop through each diagnostic. Eventually I would have to restart the language server. This change should allow it to recover during this case. I also way to explore cancellation of code action requests to see if that would provide any benefit. We should also look if we can have the parser handle an unclosed if else statement better.


razzeee:
This seems harmless enough, as long as we only do it when the root node errors, there might be helpful code actions in other cases.

PR #820, opened 2022-11-04
Fix incorrect path returned by rename-function

anagrius:
Previously the path would end with .el instead of .elm because,
the last letter was chopped off.

Ideally, the module name should be used instead of path in `name`, but this just fixes the bug.

Screenshot 2022-11-04 at 20 13 54

### DIFF
--- a/src/providers/handlers/moveRefactoringHandler.ts
+++ b/src/providers/handlers/moveRefactoringHandler.ts
@@ -12,6 +12,7 @@ import { ElmWorkspaceMatcher } from "../../util/elmWorkspaceMatcher";
 import { RefactorEditUtils } from "../../util/refactorEditUtils";
 import { References } from "../../compiler/references";
 import { TreeUtils } from "../../util/treeUtils";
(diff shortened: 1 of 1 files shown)


jmbockhorst:
Hmm, I am not able to reproduce this on Windows. Are you on a non-Windows platform?

anagrius:
No this is on mac

razzeee:
I'm also seeing this on linux

jmbockhorst:
So I would rather try to fix it properly for mac/linux.

anagrius:
Ah, not having seen it actually work I was not sure what `path` was.
I have fixed the PR to now work platform independant:

image

razzeee:
Seems fine on linux

razzeee:
@jmbockhorst can you please check windows, when you find time?

jmbockhorst:
Just tested and looks good to me.

razzeee:
Thanks for figuring this out @anagrius

PR #824, opened 2022-11-05
Add completions for destructured record patterns and ports

jmbockhorst:
Added some missing completions for destructed record patterns and ports. I was able to simplify the logic by using the existing symbol binder.

### DIFF
--- a/src/providers/completionProvider.ts
+++ b/src/providers/completionProvider.ts
@@ -955,6 +955,8 @@ export class CompletionProvider {
     const result: CompletionItem[] = [];
 
     checker.getSymbolsInScope(node, sourceFile).forEach((symbol, i) => {
+      // getSymbolsInScope returns ths symbols in order of inner scope to outer scope,
+      // so we add the index to the sort order so the variables that are "closer" to the position are sorted first
       const sortPrefix = `a${i}`;
       if (symbol.type === "Function") {
         // Only get let functions here
--- a/src/providers/completionProvider.ts
+++ b/src/providers/completionProvider.ts
@@ -953,7 +953,6 @@ export class CompletionProvider {
     range: Range,
   ): CompletionItem[] {
     const result: CompletionItem[] = [];
-    const tree = sourceFile.tree;
 
     checker.getSymbolsInScope(node, sourceFile).forEach((symbol, i) => {
       const sortPrefix = `a${i}`;
--- a/src/providers/completionProvider.ts
+++ b/src/providers/completionProvider.ts
@@ -263,7 +263,8 @@ export class CompletionProvider {
           )?.namedChildren.map((n) => n.text) ?? [];
 
         return this.getSameFileTopLevelCompletions(
-          tree,
+          checker,
+          sourceFile,
           replaceRange,
           true,
         ).filter((completion) => !exposingList.includes(completion.label));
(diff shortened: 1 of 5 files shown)


src/providers/completionProvider.ts
```diff
@@ -971,128 +947,100 @@ export class CompletionProvider {
...
+            markdownDocumentation,
+            label: symbol.name,
+            range,
+            sortPrefix,
+          }),
+        );
+
+        const annotationTypeNode =
+          TreeUtils.getTypeOrTypeAliasOfFunctionParameter(symbol.node);
+        if (annotationTypeNode) {
```
razzeee, line 1003: If my coverage report isn't lying, this branch has no test coverage. It might be nice to add this right now, to make sure we don't break something here.

src/providers/completionProvider.ts
```diff
@@ -971,128 +947,100 @@ export class CompletionProvider {
...
-                  range,
-                  sortPrefix,
-                }),
-              );
-            });
-          }
+    const tree = sourceFile.tree;
+
+    checker.getSymbolsInScope(node, sourceFile).forEach((symbol, i) => {
+      const sortPrefix = `a${i}`;
```
razzeee, line 959: Is this changing the sortPrefix? What's the reasoning behind that?

src/providers/completionProvider.ts
```diff
@@ -971,128 +947,100 @@ export class CompletionProvider {
...
+      if (symbol.type === "Function") {
+        // Only get let functions here
```
razzeee, line 961: Not sure I get why the below will only apply to let functions

razzeee:
The simplification looks very nice, great work.

src/providers/completionProvider.ts
jmbockhorst: `getSymbolsInScope` returns the symbols sorted from the inner scope to the outer scope, so I thought it might be nice if inner scope variables (closer to where you are typing) were sorted above outer scopes. Should probably add a comment.

src/providers/completionProvider.ts
jmbockhorst: Because top level functions are handled by `getSameFileTopLevelCompletions`.

src/providers/completionProvider.ts
jmbockhorst: Actually I'm not sure if this works at all. 
```elm
type alias Model = { field1 : String, field2: String }

foo : Model -> String
foo model =
   mo{-caret-}
```
I think this is trying to add completions for `model.field1` and `model.field2`. Do we want that?

src/providers/completionProvider.ts
razzeee: Yeah, that makes sense. Didn't think about that, but thought the code changed the behavior. :)

src/providers/completionProvider.ts
razzeee: I guess your right https://github.com/elm-tooling/elm-language-server/commit/719cf0fa9fee2857007986a049686bb5c1ed2d90

But it's been a hot minute since then and I have no idea, if we still want that. I would think so?

jmbockhorst: I was able to re-implement this a lot simpler using the type checker.

PR #825, opened 2022-11-05
Fix a bug with extract function parameter calculation

jmbockhorst:
This fixes a case where the extract function code action would incorrectly add parameters for inner destructured record fields. Also adds tests for other cases.

### DIFF
--- a/test/codeActionTests/extractFunctionCodeAction.test.ts
+++ b/test/codeActionTests/extractFunctionCodeAction.test.ts
@@ -18,8 +18,8 @@ foo val str =
                     prop1 + prop2 + val + field2
 
                 Nothing ->
-                               --^
                     field2 + val2
+                               --^
 
         Nothing ->
             str + val2
@@ -72,8 +72,8 @@ foo val str =
             (\\{ prop1, prop2 } ->
           --^
                 prop1 + prop2 + val + field2)
-                --^
             field1
+                --^
 
         Nothing ->
             str + val2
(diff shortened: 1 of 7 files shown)


test/utils/sourceParser.ts
```diff
@@ -159,27 +157,35 @@ export function getInvokeAndTargetPositionFromSource(source: string): TestType {
 
 export function getTargetPositionFromSource(
   source: string,
-): { position: Position; sources: { [K: string]: string } } | undefined {
+): { range: Range; sources: { [K: string]: string } } | undefined {
```
razzeee, line 160: Hrm, not your fault, but I think we don't use range at all here?

razzeee: Ignore me, those are the return values :D

test/codeActionTests/extractFunctionCodeAction.test.ts
```diff
@@ -0,0 +1,345 @@
...
+foo val str =
+    case val of
+        Just { field1, field2, field3 } ->
+            case field1 of
+          --^
+                Just { prop1, prop2 } ->
+                    prop1 + prop2 + val + field2
+
+                Nothing ->
+                               --^
```
razzeee, line 21: Are these supposed to be off by a line? I would expect them to be moved down by one?

jmbockhorst: Yeah, originally I had that, but only this worked. At the time I didn't dig into why, but I will now.

PR #898, opened 2023-02-27
Improve hover and completion type strings when there is a module prefix that uses an alias

jmbockhorst:
This fixes some cases of the module prefix being missing in a hover or completion type string when a module alias was used.


test/hoverProvider.test.ts
```diff
@@ -0,0 +1,143 @@
...
+--@ Test.elm
+module Test exposing (..)
+
+import Another exposing (foo)
+
+bar = foo
+     --^
+    `;
+
+    await testHover(source, "foo : Another.Foo");
```
razzeee, line 95: I could be wrong, but the description of this test case sounds like this should not have a module prefix?

jmbockhorst: Yeah I changed my mind on this. Originally the test was that it should not have module prefix if the current module doesn't have one, but I decided against that. I didn't change all of the test description.

jmbockhorst:
I tried using codespaces to make the edit and it messed this up. I will create a new PR.

PR #942, opened 2023-04-12
Run elm-review in a custom watch mode

jmbockhorst:
@jfmengels I've started work on trying to run elm-review using a custom runner inside the language server that can make use of watch mode. It significantly improves performance as expected, and it can show diagnostics even without file save. It is a simple implementation right now and is missing features of the normal elm-review watch mode (like files getting added, elm.json changing, etc). What do you think of this approach?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "chokidar": "^3.5.3",
+        "elm-review": "^2.9.2",
         "escape-string-regexp": "^4.0.0",
         "execa": "^5.1.1",
         "fast-diff": "^1.2.0",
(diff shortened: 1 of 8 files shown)


jmbockhorst:
Thanks for the detailed response!

On points 1 and 2, I think a public API would be great and we can work towards that. As you mentioned, the language server will still have to support multiple versions to maintain backwards compatibility, whether it is a public v1/v2 api or not for now. If we are able to use the user's version of `node-elm-review` instead of the language server's version, then we would only need to update the language server if the APIs we are using change, and support multiple version in the language server.

On point 3, ideally we would actually want to show errors for the just the current/open files first, then the rest. Since we still have the long startup time before seeing the first error on first run. 
> where I only load some files in memory (and more on demand when necessary)

I wonder if this is a similar idea.

On point 4, _theoretically_ we could use web assembly bindings to a public api of the rust or haskell runner and still run from the language server. 

Again, thanks for taking the time to think through this!

PR #955, opened 2023-04-29
Fix type inference bugs

### LINKED ISSUE #494: Type inference issues after upgrade to v2
There is an issue with type inference algo which leads to false alarms. Consider following block of code:

```
-- CONFIG

type Config data
    = Config (RequiredConfig data)

type alias RequiredConfig data =
    Data data -> String

type alias Data data =
    { data
        | windowWidth : Int
        , windowHeight : Int
    }

-- PUBLIC CONFIG

type alias PublicConfig =
    Config {}

type alias PublicRequiredConfig =
    RequiredConfig {}

publicConfig : PublicRequiredConfig -> PublicConfig
publicConfig =
    Config
```

## Expected Behavior
I expect it to compile successfuly.

## Current Behavior
LSP diagnostics report an issue at `publicConfig` function. The report is:

```
Type mismatch error.
Expected: `(PublicRequiredConfig) -> PublicConfig`
Found: `(RequiredConfig (Data {  })) -> Config (Data {  })`
```

## Possible Solution
It seems that what LSP finds is wrong. Because it should be `(RequiredConfig {  }) -> Config {  }` instead of `(RequiredConfig (Data {  })) -> Config (Data {  })`.

Run LSP diagnostics on provided block of the code.

## Context
The issue reports false alarms.

## Your Environment

* Version used: v2.0.1
* Editor used: Emacs
* Environment name and version (e.g. node.js 5.4): v14.15.1
* Operating System and version: MacOS 10.15.7

### LINKED ISSUE #620: Incorrect type inference for records applied to lists of functions
SSCCE:
```elm
func r =
    List.map ((|>) r) [ \{ foo } -> foo, \{ bar } -> bar ]
```
## Expected Behavior

```elm
func : { a | foo : b, bar : b } -> List b
func r =
    List.map ((|>) r) [ \{ foo } -> foo, \{ bar } -> bar ]
```

## Current Behavior
```elm
func : { a | foo : b } -> List b
func r =
    List.map ((|>) r) [ \{ foo } -> foo, \{ bar } -> bar ]
```

## Additional Observations

Only the **first** function (by order in the list) will be used for the type inference, no matter the number.  A few more examples with their incorrect inferences are provided here to illustrate:

```elm
f1 : { a | foo : b } -> List b
f1 r =
    List.map ((|>) r)
        [ \{ foo } -> foo
        , \{ bar } -> bar
        ]


f2 : { a | bar : b } -> List b
f2 r =
    List.map ((|>) r)
        [ \{ bar } -> bar
        , \{ foo } -> foo
        ]


moreThanTwo : { a | bar : b } -> List b
moreThanTwo r =
    List.map ((|>) r)
        [ \{ bar } -> bar
        , \{ foo } -> foo
        , \{ baz } -> baz
        ]
```

My assumption is that only the type of the **first** element in a list is looked at to determine the type of the list, but this is insufficient if the first element has the inferred type of an extensible record.

Interestingly enough, it looks like more than one element is checked for other cases where the first might be ambiguous, e.g.

```elm
number : Float -> List Float
number i =
    List.map ((|>) i)
        [ (+) 1
        , (/) 2
        ]
```

is correctly inferred as `Float` not as `number`, which is what would be determined from only the first element.

## Context

This seems contrived, but it came up while writing tests, using [`Expect.all`](https://package.elm-lang.org/packages/elm-explorations/test/latest/Expect#all), e.g.

```elm
import Expect

expect =
    Expect.all
        [ \{ foo } -> Expect.equal 1.1 foo
        , \{ bar } -> Expect.equal "2" bar
        ]
```

will infer as 

```elm
expect : { a | foo : Float } -> Expect.Expectation
expect =
    Expect.all
        [ \{ foo } -> Expect.equal 1.1 foo
        , \{ bar } -> Expect.equal "2" bar
        ]
```

rather than the correct

```elm
expect : { a | foo : Float, bar : String } -> Expect.Expectation
expect =
    Expect.all
        [ \{ foo } -> Expect.equal 1.1 foo
        , \{ bar } -> Expect.equal "2" bar
        ]
```

## Your Environment

* Version used: elmtooling.elm-ls-vscode v2.2.0
* Editor used: VS Code v1.57.1 
* Operating System and version: NixOS

### LINKED ISSUE #634: Treatment of constrained type variables in type aliases differs from Elm compiler
Constrained type variables (e.g. `appendable`) are respected in type aliases by the language server.  For example, the language server flags:
```elm
type alias A appendable =
    appendable

a : A Int
a =
    1
```
as an error, because `Int` is not `appendable`.  The Elm compiler, however, does **not** enforce constrained type variables for type aliases, so the above is valid Elm code.

There is a more detailed discussion of the compiler's exact treatment of constrained type variables up here: elm/compiler#2225

TL;DR: They are respected for record constructors, but otherwise not for type aliases, and they are never obeyed at the type level, even for custom types.

## Expected Behavior

Type errors should only be reported where the compiler would report them.

## Current Behavior

The language server is more strict in its understanding of constrained type variables.

The above code snippet is a good, simple example.

## Your Environment

* Version used: v2.2.1
* Editor used: VS Code 1.57.1

### LINKED ISSUE #678: Extensible records aren't typechecked properly
Extensible records aren't typechecked properly. The language server throws errors where the the elm compiler does not, and vice versa.

I found two different kinds of errors do let me know if you'd like me to split into multiple issues.

### Unbound type variables
The following two examples should error because unbound type variables are used:
```elm
type Example = Example { a | foo : Float }
```

```elm
type alias Example = { a | foo : Float }
```

An error similar to the elm compiler one should be thrown:
```
The `Example` type alias uses an unbound type variable `a` in its definition:

8| type alias Example = { a | foo : Float }
                          ^
You probably need to change the declaration to something like this:

    type alias Example a = ...

Why? Well, imagine one `Example` where `a` is an Int and another where it is a
Bool. When we explicitly list the type variables, the type checker can see that
they are actually different types.
```

### Incorrect type inference
This shouldn't error as elm compiler typechecks it fine:
```elm
type Example a = Example { a | foo : Float }

example : Example {}
example = Example { foo = 0 }
```

### Unbound type variables
The following two examples throw no errors, despite use of unbound type variables.
```elm
type Example = Example { a | foo : Float }
```

```elm
type alias Example = { a | foo : Float }
```

### Incorrect type inference
The following example...
```elm
type Example a = Example { a | foo : Float }

example : Example {}
example = Example { foo = 0 }
```
...shows the following error...
```
Type mismatch error.
Expected: `Example {  }`
Found: `Example { foo : Float }`Elm
```
...when it should typecheck fine.

## Context
I have a project that compiles fine, but when trying to move to using VSCode errors are shown for code that compiles fine. This project is open source, so here is an example [file with errors](https://github.com/RoganMurley/GALGAGAME/blob/master/client/elm/Room/Connected/GameState/PlayState/Model/Hover.elm).

## Your Environment

* Version used: 2.4.0
* Editor used: Elm Plugin for Visual Studio Code 
* Environment name and version : node v12.22.5
* Operating System and version: Ubuntu 21.10

### LINKED ISSUE #689: Type inference error when using a generic type alias
`RangeFilterData` is getting bound to `RangeFilterData Float` and then failing when trying to decode for a date.
``` elm
case ... of 
    "NumericRangeFilter" ->
        (Decode.succeed RangeFilterData
            |> required "min" (Decode.nullable Decode.float)
            |> required "max" (Decode.nullable Decode.float)
        )
            |> Decode.map NumericRangeFilter

    "DateRangeFilter" ->
        (Decode.succeed RangeFilterData
            |> required "min" (Decode.nullable dateDecoder)
            |> required "max" (Decode.nullable dateDecoder)
        )
            |> Decode.map DateRangeFilter


type alias RangeFilterData data =
    { min : Maybe data
    , max : Maybe data
    }

type FilterType
    = NumericRangeFilter (RangeFilterData Float)
    | DateRangeFilter (RangeFilterData Date)


```

jmbockhorst:
Fixes #689, fixes #634, fixes #620, fixes #678, fixes #494.

### DIFF
--- a/test/performance.ts
+++ b/test/performance.ts
@@ -70,10 +70,10 @@ export async function runPerformanceTests(uri: string): Promise<void> {
     times[name] = (times[name] ?? 0) + time;
   }
 
-  function addPerfTime(name: string, time: PerformanceTimer): void {
-    times[name] = (times[name] ?? 0) + time.total;
-    time.reset();
-  }
+  // function addPerfTime(name: string, time: PerformanceTimer): void {
+  //   times[name] = (times[name] ?? 0) + time.total;
+  //   time.reset();
+  // }
 
   const numTimes = 10;
 
--- a/test/performance.ts
+++ b/test/performance.ts
@@ -23,6 +23,7 @@ import {
 } from "../src/cancellation";
 import { randomBytes } from "crypto";
 import { createProgramHost } from "./utils/sourceTreeParser";
+import { PerformanceTimer } from "../src/compiler/typeExpression";
 
 container.register("Connection", {
   useValue: {
@@ -67,6 +68,11 @@ export async function runPerformanceTests(uri: string): Promise<void> {
   const times: { [name: string]: number } = {};
   function addTime(name: string, time: number): void {
     times[name] = (times[name] ?? 0) + time;
+  }
+
+  function addPerfTime(name: string, time: PerformanceTimer): void {
+    times[name] = (times[name] ?? 0) + time.total;
+    time.reset();
   }
 
   const numTimes = 10;
--- a/src/compiler/typeCache.ts
+++ b/src/compiler/typeCache.ts
@@ -8,18 +8,23 @@ type CacheKey =
   | "PACKAGE_TYPE_ANNOTATION"
   | "PACKAGE_TYPE_AND_TYPE_ALIAS"
   | "PACKAGE_VALUE_DECLARATION"
+  | "PACKAGE_UNION_VARIANT"
   | "PROJECT_TYPE_ANNOTATION"
   | "PROJECT_TYPE_AND_TYPE_ALIAS"
-  | "PROJECT_VALUE_DECLARATION";
+  | "PROJECT_VALUE_DECLARATION"
+  | "PROJECT_UNION_VARIANT";
 
 export class TypeCache {
   private packageTypeAnnotation: SyntaxNodeMap<SyntaxNode, InferenceResult>;
   private packageTypeAndTypeAlias: SyntaxNodeMap<SyntaxNode, InferenceResult>;
   private packageValueDeclaration: SyntaxNodeMap<SyntaxNode, InferenceResult>;
+  private packageUnionVariant: SyntaxNodeMap<SyntaxNode, InferenceResult>;
   private projectTypeAnnotation: SyntaxNodeMap<SyntaxNode, InferenceResult>;
   private projectTypeAndTypeAlias: SyntaxNodeMap<SyntaxNode, InferenceResult>;
   private projectValueDeclaration: SyntaxNodeMap<SyntaxNode, InferenceResult>;
+  private projectUnionVariant: SyntaxNodeMap<SyntaxNode, InferenceResult>;
   private declarationAnnotations: MultiMap<number, SyntaxNode>;
+  private typeUnionVariants: MultiMap<number, SyntaxNode>;
 
   constructor() {
     this.packageTypeAnnotation = new SyntaxNodeMap<
@@ -34,6 +39,7 @@ export class TypeCache {
       SyntaxNode,
       InferenceResult
     >();
+    this.packageUnionVariant = new SyntaxNodeMap<SyntaxNode, InferenceResult>();
     this.projectTypeAnnotation = new SyntaxNodeMap<
       SyntaxNode,
       InferenceResult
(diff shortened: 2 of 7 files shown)


test/performance.ts
```diff
@@ -69,6 +70,11 @@ export async function runPerformanceTests(uri: string): Promise<void> {
     times[name] = (times[name] ?? 0) + time;
   }
 
+  function addPerfTime(name: string, time: PerformanceTimer): void {
```
razzeee, line 73: This seems unused?

Issue #963, opened 2023-05-05
Support running the server in the web


jmbockhorst:
We should be able to support some basic things in a non node environment.

PR #964, opened 2023-05-08
Support virtual file systems

jmbockhorst:
This is the first step of #963. 
Corresponding vscode client PR: https://github.com/elm-tooling/elm-language-client-vscode/pull/560.

There is a lot going on here, so I will try to explain the main changes. 

- Initial indexing is moved out of the `onInitialize` request and will now happen after `onInitialized`. This is necessary because in a virtual workspace we need to send a custom request (`ReadFileRequest`) to the client to read files, but we can't send requests until it `onInitialize` is completed. This is consistent with other language server extensions that I have looked at. In the future I would like to explore if we can move some of the up front indexing to be done more lazily.
- The client is required to send `elmJsonFiles` for it to support virtual workspaces, because of the same reason above of the server not being able to send a request during `onInitialize`. Theoretically this could even be moved until after initialization.
- Because of that, `ElmWorkspaceMatcher` handlers need to be async so that they can wait until the program is initialized.
- If it is a virtual workspace, we can't run any cli tools, so we can't rely on `elm make` to download package dependencies. 
- Thus, in a virtual workspace we use a similar process that the elm cli uses. We check the package website for versions and then load the sources directly from github. Then we cache those packages in our own cache on the file system (`~/.elm/elm-language-server/packages`) so we don't have to download them in the future.
- There is a special `TextDocumentContentProvider` so that we can show the "virtual" package files that don't exist on disk (although they do in the cache)
- I started splitting up node vs web, but the rest of that work will happen in a separate PR.

The experience should feel unchanged for a normal file system workspace.


razzeee:
I will wait for tests to pass, before reviewing

jmbockhorst:
This should be ready to review now. See updated description above.

PR #1043, opened 2023-08-22
Fix case statement constant binding inference

micahhahn:
I noticed recently that `elm-language-server` would infer the wrong types for a simple case statement like the following: 

```elm
func x =
    case x of 
        "a" ->
            1
 
        _ ->
            2
```

The elm compiler infers a type of `String -> number`, but `elm-language-server` suggests `a -> number`. 

The fix seemed simple enough (I think?!?).  Making sure that the variable is assignable to the constants type did the trick.

### DIFF
--- a/src/compiler/typeInference.ts
+++ b/src/compiler/typeInference.ts
@@ -1793,22 +1793,22 @@ export class InferenceScope {
         this.isAssignable(pattern, ty, TUnit, undefined, true);
         break;
       case "StringConstant":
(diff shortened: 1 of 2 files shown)


razzeee:
Looks straight forward enough, but @jmbockhorst is the expert here :)

micahhahn:
Bump to bring this back to visibility... I think this change is straightforward - would be happy to jump on a zoom and discuss

Issue #1052, opened 2023-09-01
tree-sitter-elm.wasm not found


jweir:
Starting vim with elm-language-server 2.7.1 I get the following error:
```
[coc.nvim] Request initialize failed with message: ENOENT: no such file or directory, open '../../../../../../opt/homebrew/lib/node_modules/@elm-tooling/
[coc.nvim] Server languageserver.elmLS failed to start: Error: Request initialize failed with message: ENOENT: no such file or directory, open '../../../
../../../opt/homebrew/lib/node_modules/@elm-tooling/elm-language-server/out/common/tree-sitter-elm.wasm'
```

## Current Workaround

I adjusted the path to tree-sitter-elm to the below, but this probably isn't a great solution.

```
        const pathToWasm = (_b = initializationOptions.treeSitterElmWasmUri) !== null && _b !== void 0 ? _b : Path.relative(process.cwd(), Path.join(__dirname, '../../out/', "tree-sitter-elm.wasm"));
```

## Your Environment

* Version used: 2.7.1
* Editor used: vim + coc 
* Environment name and version (e.g. node.js 5.4): node v18.16.1
* Operating System and version: MacOS 13.4.1

razzeee:
Could be this change https://github.com/elm-tooling/elm-language-server/commit/1dd8d658a7ac79e50136d98242abb27af6ee87c6#diff-1f963b5ae3613de37b5354a3da594cdef878e0715ad0eb3896bd701e2511b273R60

Chocrates:
Same issue here with a relatively new version of NVIM.  Using Packer and Mason and did a `npm install -g @elm-tooling/elm-language-server` though not sure that was needed.
Gonna try the workaround.
Where did you put this workaround @jweir?

jweir:
@Chocrates  I edited `out/common/index.js` line 74 under `/opt/homebrew/lib/node_modules/@elm-tooling/elm-language-server/` – your location may vary.

Chocrates:
Thanks that worked for me @jweir

dtrckd:
Same error here. I paste the stack I got in neovim as it is slightly different from @jweir, in case it could helps

 ``
Error executing vim.schedule lua callback: /snap/nvim/2809/usr/share/nvim/runtime/lua/vim/lsp.lua:1309: RPC[Error] code_name = InternalError, message = "Request initialize failed with message: ENOENT: no such file
[...]

PR #1070, opened 2023-09-22
fix: use the correct path to tree-sitter-elm.wasm

seanstrom:
## Description

* This fix is intended to solve #1052 
* The cause of the issue may be related to a recent refactor where the logic for building the file path to the `tree-sitter-elm.wasm` file had silently changed.
* There could be other solutions for this issue, but this PR attempts to simply correct the issue, but perhaps there's a better way for sharing the path to the `out` directory.

### DIFF
--- a/src/common/index.ts
+++ b/src/common/index.ts
@@ -20,6 +20,7 @@ import { Settings } from "./util/settings";
 import { TextDocumentEvents } from "./util/textDocumentEvents";
 import { IFileSystemHost, InitializationOptions } from "./types";
 import { URI, Utils } from "vscode-uri";
+import { outDir } from "../directories";
 
 export function startCommonServer(
   connection: Connection,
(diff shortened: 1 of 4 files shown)


jmbockhorst:
Yeah I guess this is fine, but we will just need to adjust the vscode client because this breaks it.

seanstrom:
Hey thanks for the feedback, 

After looking at the VSCode extension, I can see what you mean by my original change causing the VSCode client to break. Seems like the VSCode client bundles the code for the extension, so `Path.resolve(__dirname, "..")` won't point to the correct place and `__dirname` should be used instead.

With that in mind, I've made another change where I created a `src/directories.ts` module. That module basically exports `__dirname` and is used inside `common/index.ts`. 

The goal here is to rely on the placement of the file in the `src` directory to resolve the correct `__dirname` for the Neovim client, while also supporting the VSCode client when the code is bundled.

Let me know what you think 🙏

jmbockhorst:
Yep, this looks good to me.

jmbockhorst:
Looks like it is failing prettier formatting.

seanstrom:
Just pushed up a formatting fix for that 👍 
Oh and btw should I squash commit my stuff?

jmbockhorst:
You can or I can when I merge it.

seanstrom:
Okay I just squashed 🙌

PR #1131, opened 2023-12-15
Fix bug when the client sends multiple contentChanges at once

### LINKED ISSUE #878: Yet More Erratic / Broken Behaviour
When using elm-language-server version `2.6.0` in neovim, **both with COC and with native LSP support**, the project will randomly be covered in errors. Other times (more frequently), nearly every top level definition (including `main`) is flagged with an `Unused pattern variable` warning. Typing in or saving the file seems to randomly both induce and relieve this issue, and its occurrence seems to be more frequent on larger projects.

I've attached two screenshots, the first with COC and the second with the native LSP. In addition, the actual installation of elm-language-server varied between the two (the first done globally on my system through npm, the second handled by mason), so it's unlikely that was the root cause.

This is in a project that compiles and runs without error:

Screen Shot 2023-01-17 at 1 08 42 AM
Screen Shot 2023-01-17 at 1 18 56 AM

As this seemed similar to #503 and #598 I tried all proposed solutions there, none seeming to work. Sorry for this being a bit vague, if there's more specific information that would be helpful please let me know how to provide it!

Cheers

peret:
On a textDocument/didChange notification, if the contentChanges contain multiple entries, each of them needs to be handled as if all previous changes were already applied. When we don't, the calculation of the change's byte offsets will be wrong, which can lead to a syntax tree that produces parsing errors where the actual file does not have any issues.
To solve this, we now apply the text of each change record to the full file text and use the result when applying the next change of the contentChanges array.

I think this will fix #878  (see issue comments for details).

### DIFF
--- a/src/common/providers/astProvider.ts
+++ b/src/common/providers/astProvider.ts
@@ -107,25 +107,28 @@ export class ASTProvider {
     let hasContentChanges = false;
     if ("contentChanges" in params && params.contentChanges) {
       hasContentChanges = true;
+      let currentText = tree?.rootNode.text;
       for (const change of params.contentChanges) {
-        if ("range" in change) {
-          tree?.edit(this.getEditFromChange(change, tree.rootNode.text));
-        } else {
-          const currentText = tree?.rootNode.text;
-          if (currentText) {
-            const regex = new RegExp(/\r\n|\r|\n/);
-            const lines = currentText.split(regex);
-            const range = {
-              start: { line: 0, character: 0 },
-              end: { line: lines.length, character: 0 },
-            };
-            tree?.edit(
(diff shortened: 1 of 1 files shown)


peret:
I am not sure whether this would have any noticeable performance impact, especially since most editors will send this notification on every keystroke?
Since only events with more than one contentChanges entry are of interest here, it would be pretty easy to add a check and only call `applyChangeToDocument()` when needed.

tankorsmash:
Necro'ing the thread to say thank you for this patch! I updated from 2.7.1 to 2.8.0 and the LSP doesn't seem to be getting it wrong anymore. Thank you!

PR #1135, opened 2023-12-21
Add elm-format as dev dependency so tests can be run locally without a global install of elm-format

Qluxzz:
Context: https://github.com/elm-tooling/tree-sitter-elm/pull/145#pullrequestreview-1791679811

Adds elm-format as a dev dependency to allow test suite to run successfully without the need for a global install of elm-format.

This also makes it so the version of elm-format is stable between test runs.


razzeee:
Not sure why macos is failing, but we'll drop node 16 soon hopefully

PR #1158, opened 2024-01-26
Round progress percentage to match the spec

w0rm:
@absynce and I are trying to integrate `elm-language-server` with [Zed](https://github.com/zed-industries/zed) and running into the following error:

```
[2024-01-25T20:44:07Z ERROR lsp] error deserializing $/progress request: Error("data did not match any variant of untagged enum ProgressParamsValue", line: 1, column: 127), message: "{\"token\":\"36abce40-63b2-4fa1-8464-030400636f26\",\"value\":{\"kind\":\"report\",\"percentage\":100.00000000000155,\"message\":\"Indexing\"}}"
```

According to [the lsp specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workDoneProgress) the `percentage` should be a uinteger. However, the actual value is a float. This works just fine if the receiving end is not strict about this, but the rust lsp implementation [assumes `percentage` to be a u32](https://docs.rs/lsp-types/latest/lsp_types/struct.WorkDoneProgressReport.html).

This PR fixes that by rounding the progress.

### DIFF
--- a/src/common/server.ts
+++ b/src/common/server.ts
@@ -169,7 +169,9 @@ export class Server implements ILanguageServer {
             const avgIndexed =
               all.reduce((sum, { indexedPercent }) => sum + indexedPercent, 0) /
               all.length;
-            progress.report(avgIndexed, "Indexing");
+            // Workaround: Rounding percentage to uint as per LSP spec
+            // https://github.com/microsoft/vscode-languageserver-node/issues/1412
+            progress.report(Math.round(avgIndexed), "Indexing");
           }),
         ),
     );
--- a/src/compiler/program.ts
+++ b/src/compiler/program.ts
@@ -407,7 +407,9 @@ export class Program implements IProgram {
             .then((progress) => {
               progress.begin("Restarting Elm Language Server", 0);
               this._initializeProgressCallback = (percent: number): void => {
-                progress.report(percent, `${percent.toFixed(0)}%`);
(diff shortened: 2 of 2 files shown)


razzeee:
That might be in part a typescript problem, but I guess the library should just do the rounding ideally? https://github.com/microsoft/vscode-languageserver-node/blob/c0982bf544bab9600df223146dd75c93c9004dc3/server/src/common/progress.ts#L22

Would be great to get this fixed upstream - can you at least create an issue there?

w0rm:
@razzeee good point! I am trying to make sense of the library and seeing that the API to report progress is different. They seem to expect an object argument instead of percentage and message.

w0rm:
Found it, https://github.com/microsoft/vscode-languageserver-node/blob/main/server/src/common/progress.ts#L57 this is probably where it should be rounded, just before sending the notification.

w0rm:
@razzeee I just opened https://github.com/microsoft/vscode-languageserver-node/issues/1412 

Would you be up for merging this change as a workaround or rather wait for it to be fixed upstream?

razzeee:
I'm fine with merging this as a workaround, but not sure if we should do a release now?

w0rm:
@razzeee just updated the comment to mention that this is a workaround with a link to the upstream issue. 

Yeah no need to cut a release right now. There is another issue I'd like to research, maybe another change would be necessary.

PR #1159, opened 2024-01-26
Fix the issue with the missing diagnostic.data

w0rm:
When integrating with Zed, I noticed the following error coming from elm-language-server:

```
[2024-01-25T20:47:42Z ERROR util] crates/editor/src/editor.rs:3506: Request textDocument/codeAction failed with message: Cannot read properties of undefined (reading 'code')
```

I narrowed it down to the code where we get `registrations` by `diagnostic.data.code`. Somehow, there is a diagnostic, that doesn't have the `data` property. It is of the following shape: `{"range":{"start":{"line":0,"character":0},"end":{"line":0,"character":0}},"severity":1,"message":""}`.

Expecting and handing the missing `data`, similar to how we do it in [hasElmReviewFixes](https://github.com/elm-tooling/elm-language-server/blob/main/src/common/providers/diagnostics/elmReviewDiagnostics.ts#L28) made the error go away.

cc @absynce

### DIFF
--- a/src/common/providers/codeActionProvider.ts
+++ b/src/common/providers/codeActionProvider.ts
@@ -293,6 +293,12 @@ export class CodeActionProvider {
     // and the fix all code action for that error if there are other diagnostics with
     // the same error code
     (<IDiagnostic[]>params.context.diagnostics).forEach((diagnostic) => {
+      // We have type casted diagnostic to `IDiagnostic`, however the original
+      // type is `Diagnostic`, the `.data` property may be missing in some cases
+      if (!diagnostic.data) {
+        return;
+      }
+
       const registrations =
         CodeActionProvider.errorCodeToRegistrationMap.getAll(
           diagnostic.data.code,


razzeee:
Finding the actual origin would be nice, but I guess very hard due to the data :/

w0rm:
@razzeee yeah it's quite hard to make sense of this. The lifetime of a `Diagnostic` / `IDiagnostic` is not clear to me.

w0rm:
@razzeee I noticed that such diagnostic is being sent from `Zed`, so there is nothing we can do about it rather than filter it:

```
{"jsonrpc":"2.0","id":68,"method":"textDocument/codeAction","params":{"textDocument":{"uri":"file:///Users/unsoundscapes/Src/Github/elm-pool/src/Game.elm"},"range":{"start":{"line":185,"character":33},"end":{"line":185,"character":33}},"context":{"diagnostics":[{"range":{"start":{"line":0,"character":0},"end":{"line":0,"character":0}},"severity":1,"message":""}],"only":["","quickfix","refactor","refactor.extract","source"]}}}
```

w0rm:
@razzeee I also realised that the settings are being overridden. `Zed` was sending an empty object that resulted in complete settings reset.

According to the documentation, only the changed settings have to be included. I changed the code to rather extend the existing settings object.

razzeee:
I think this is fine then

w0rm:
So what I think actually happens in the case of configuration change:

```
// Zed receives the request from elmLS
{"jsonrpc":"2.0","id":8,"method":"workspace/configuration","params":{"items":[{"section":"elmLS"}]}}
// Zed responds with empty value
{"jsonrpc":"2.0","id":8,"result":[null],"error":null}
```
According to https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#workspace_configuration 

>  If the client can’t provide a configuration setting for a given scope then `null` needs to be present in the returned array.

I am really not sure why `onDidChangeConfiguration(params)` is triggered in such case with settings set to `{}`.

w0rm:
@razzeee maybe we can change this to check if the settings is an empty object then do nothing?

w0rm:
@razzeee I removed the change that merges the configuration because this is questionable. I may be able to find a way to fix this on the Zed's end. 

I think that the diagnostics check can be merged though!

PR #1164, opened 2024-02-02
Remove associated data when removing type value

Qluxzz:
Currently when trying to remove an unused type variant with associated data using the code action `Remove unused value constructor` the result is:

Before:
```elm
type Foo
    = Bar
    | Baz Int
```

After:
```elm
type Foo
    = Bar Int
```
It removed the type variant `Baz` but not the associated data but instead moved that to the previous type constructor

The following changes fixes that and also adds tests to verify that `Remove unused value constructor` in general works as expected.


razzeee:
Thank you

PR #1175, opened 2024-02-17
Fix Add missing case branches action adding missing branches outside of parenthesis

Qluxzz:
Currently if you have the following:
```elm
func a =
    (case a of
        1 ->
            ""
        
    )
        |> Debug.log "test"
```

And then use the `Add missing case branches` action, the result is:

```elm
func a =
    (case a of
        1 ->
            ""
        
    )

        _ ->
            Debug.todo "branch '_' not implemented"
        |> Debug.log "test"
```

The following changes fixes that by since the case_of_expr's endPosition includes the ending ) we instead try to find the last case_of_branch, and take its position as the insert position.


razzeee:
Thank you

PR #1178, opened 2024-02-24
Include type alias field line comment in hover info

Qluxzz:
Currently when hovering on a type alias field this is what's shown
image

The line comment is not included in the hover text.

The following changes makes it so if a line comment is present for a given type alias field, it is included in the hover info as follows:

image


razzeee:
Thank you

PR #1179, opened 2024-02-25
Fix add missing case branches not being prefixed correctly

Qluxzz:
Add missing case branches does not include prefix for variant
```elm
test : Http.Response () -> ()
test response =
    case response of
        Http.GoodStatus_ _ _ ->
            ()
```

After using code action `Add missing case branches`:

```elm
test : Http.Response () -> ()
test response =
    case response of
        Http.GoodStatus_ _ _ ->
            ()

        BadUrl_ _ ->
            Debug.todo "branch 'BadUrl_ _' not implemented"

        Timeout_ ->
            Debug.todo "branch 'Timeout_' not implemented"

        NetworkError_ ->
            Debug.todo "branch 'NetworkError_' not implemented"

        BadStatus_ _ _ ->
            Debug.todo "branch 'BadStatus_ _ _' not implemented"
```
Which fails to compile since the variants without prefix are not available in the scope of the file.

### DIFF
--- a/src/common/providers/codeAction/addMissingCaseBranchesCodeAction.ts
+++ b/src/common/providers/codeAction/addMissingCaseBranchesCodeAction.ts
@@ -68,7 +68,9 @@ function getEdits(params: ICodeActionParams, range: Range): TextEdit[] {
 
     const edit = PatternMatches.missing(patterns, params.program).reduce(
       (edit, missing) => {
-        if (prefix) missing = `${prefix}.${missing}`;
+        if (prefix) {
+          missing = `${prefix}.${missing}`;
+        }
 
         return `${edit}\n\n${branchIndent}${missing} ->\n${branchExprIndent}Debug.todo "branch '${missing}' not implemented"`;
       },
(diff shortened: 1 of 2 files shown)


src/common/providers/codeAction/addMissingCaseBranchesCodeAction.ts
```diff
@@ -58,10 +58,20 @@ function getEdits(params: ICodeActionParams, range: Range): TextEdit[] {
...
+      .slice(0, -1) // Don't take the last one since that's the variant
+      .map((x) => x.text)
+      .join(".");
+
     const edit = PatternMatches.missing(patterns, params.program).reduce(
-      (edit, missing) =>
-        edit +
-        `\n\n${branchIndent}${missing} ->\n${branchExprIndent}Debug.todo "branch '${missing}' not implemented"`,
+      (edit, missing) => {
+        if (prefix) missing = `${prefix}.${missing}`;
```
razzeee, line 71: Can you change this to be braced? I'm not a big fan of  the braceless `if` statement

Qluxzz: I have added braces

PR #1472, opened 2025-11-03
Diagnostic providers should respect elm-analyse excluded files

bonega:
This PR ensures that all diagnostic providers respect the `excludedPaths` setting from `elm-analyse.json` configuration file.

Key Changes:

- Applied exclusion checks across all diagnostic types (elm-make, syntactic, semantic, suggestions, and elm-ls)
- Files matching excluded paths now skip compilation and diagnostic checks entirely
- This ensures consistent behavior across all diagnostic providers when users configure paths to exclude from analysis.

Use case:
In our current codebases we have a lot of git submodules where we don't care if they are following code convention etc, in this way we can silence them and save a few cycles.

### DIFF
--- a/test/jest.setup.ts
+++ b/test/jest.setup.ts
@@ -29,6 +29,9 @@ container.registerSingleton<IElmAnalyseJsonService>(
     public getElmAnalyseJson(workspacePath: string): IElmAnalyseJson {
       return { checks: { SingleFieldRecord: true } };
     }
+    public isFileExcluded(fileUri: string, workspacePath: string): boolean {
+      return false;
+    }
   },
 );
 container.register(ElmMakeDiagnostics, {


src/common/providers/diagnostics/elmMakeDiagnostics.ts
```diff
@@ -204,7 +209,11 @@ export class ElmMakeDiagnostics {
       ? forestFiles
       : forestFiles.concat(sourceFile);
 
-    const projectFiles = allFiles.filter((file) => !file.isDependency);
+    const projectFiles = allFiles.filter(
+      (file) =>
+        !file.isDependency &&
```
razzeee: This feels, like it should be an `||`

bonega: I think it should be `&&`

include file if: 
* it is not a dependency 
* it is not excluded

Both should apply

PR #1494, opened 2026-08-13
fix: deduplicate transitive module references

### LINKED ISSUE #1493: getImportingModules mutates the import graph, causing textDocument/references to exceed the call stack
Disclosure: report generated by gpt-5.6-sol however I have confirmed the proposed change works with a nix patch in my codebase

# Description

Requesting references for a record field declared in a widely imported module fails with `Request textDocument/references failed with message: Maximum call stack size exceeded`

This occurs in elm-language-server 2.8.0 and is still present on main.

# Environment

- elm-language-server: 2.8.0
- Elm: 0.19.1
- Node.js: 22.22.2
- Platform: macOS ARM64
- Client: Emacs lsp-mode
- Also reproduced directly over LSP without Emacs

# Reproduction

In a sufficiently large Elm application:

1. Declare a record type alias in a central module.
2. Import that module through a branching/diamond-shaped module graph.
3. Send textDocument/references for one of the record-field declarations.

The project compiles successfully and contains no import cycle.

Example request:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "textDocument/references",
  "params": {
    "textDocument": {
      "uri": "file:///path/to/project/src/Shared/Model.elm"
    },
    "position": {
      "line": 111,
      "character": 10
    },
    "context": {
      "includeDeclaration": true
    }
  }
}
```

# Actual behavior

The server returns:

```json
{
  "code": -32603,
  "message": "Request textDocument/references failed with message: Maximum call stack size exceeded"
}
```

The relevant stack frames are:

```
RangeError: Maximum call stack size exceeded
    at typeChecker.js:224
    at Array.forEach
    at getAllDescendantImports
    at typeChecker.js:224
    at Array.forEach
    at getAllDescendantImports
```

# Root cause

`getImportingModules` retrieves an array directly from the cached `importModuleGraph` and then appends transitive importers to that same array

```ts
const getAllDescendantImports = (source: ISourceFile): ISourceFile[] => {
  const imports = importModuleGraph.get(source.uri) ?? [];

  imports.slice(0).forEach((directImport) => {
    imports.push(...getAllDescendantImports(directImport));
  });

  return imports;
};
```

This has two effects:

1. It permanently mutates the cached direct-importer graph.
2. It retains and repeatedly expands duplicate paths through diamond-shaped graphs.

In the application where I reproduced this, the field’s module had:

- 43 direct importers
- 95 unique transitive importers
- 232,703 entries after duplicate expansion

The final `imports.push(...largeArray)` exceeds V8’s argument/call-stack limit.

# Expected behavior

The server should traverse each importing module at most once and return the record-field references without error.

# Proposed fix

Use a visited set and avoid mutating `importModuleGraph`

```ts
const importingModules: ISourceFile[] = [];
const visited = new Set<string>([sourceFile.uri]);
const modulesToVisit = [sourceFile];

for (let index = 0; index < modulesToVisit.length; index++) {
  const source = modulesToVisit[index];

  (importModuleGraph.get(source.uri) ?? []).forEach((importingModule) => {
    if (!visited.has(importingModule.uri)) {
      visited.add(importingModule.uri);
      importingModules.push(importingModule);
      modulesToVisit.push(importingModule);
    }
  });
}

return importingModules;
```

jasonatwell:
disclosure: generated by gpt-5.6-sol

Avoid mutating the cached import graph when finding record field references, preventing duplicate results and call stack overflows.

I have confirmed the proposed change works with a nix patch in my codebase

closes https://github.com/elm-tooling/elm-language-server/issues/1493


razzeee:
Did you test this?

src/compiler/typeChecker.ts
```diff
@@ -369,17 +369,23 @@ export function createTypeChecker(program: IProgram): TypeChecker {
       return importModuleGraph.get(sourceFile.uri) ?? [];
     }
 
-    const getAllDescendantImports = (source: ISourceFile): ISourceFile[] => {
-      const imports = importModuleGraph.get(source.uri) ?? [];
-
-      imports.slice(0).forEach((directImport) => {
-        imports.push(...getAllDescendantImports(directImport));
+    const importingModules: ISourceFile[] = [];
```
razzeee: I would prefer this whole thing to be written in a more functional and non mutating way, if possible.

jasonatwell:
> Did you test this?

@razzeee yes I did, I just updated the PR description to state "I have confirmed the proposed change works with a nix patch in my codebase"

src/compiler/typeChecker.ts
jasonatwell: > I would prefer this whole thing to be written in a more functional and non mutating way, if possible.

I agree, but I figured since this is a performance patch the mutable version is likely more performant. To check that assumption I had an LLM create a test and the original code is about 90x more performant than its refactored functional code

### test results

Conclusion: the original mutable breadth-first loop is about 90× faster on average for this workload.

Method: ran the stress benchmark 10 times in fresh Jest processes. Each run performs 20 traversals of a 576-module graph with duplicate edges and a cycle.

```
│ Run │ Mutable loop │ Immutable traversal │
├─────┼──────────────┼─────────────────────┤
│ 1   │ 0.87 ms      │ 78.67 ms            │
│ 2   │ 1.05 ms      │ 83.15 ms            │
│ 3   │ 0.87 ms      │ 82.33 ms            │
│ 4   │ 0.96 ms      │ 84.67 ms            │
│ 5   │ 0.88 ms      │ 82.38 ms            │
│ 6   │ 0.97 ms      │ 83.39 ms            │
│ 7   │ 0.86 ms      │ 82.98 ms            │
│ 8   │ 0.87 ms      │ 83.14 ms            │
│ 9   │ 0.96 ms      │ 86.17 ms            │
│ 10  │ 0.88 ms      │ 83.65 ms            │
```

### refactored functional code

```ts
export function getTransitiveImportingModules<T extends { uri: string }>(
  importModuleGraph: ReadonlyMap<string, readonly T[]>,
  sourceFile: T,
): T[] {
  const findTransitiveImporters = (
    modulesToVisit: readonly T[],
    visited: ReadonlySet<string>,
  ): T[] => {
    const [source, ...remainingModules] = modulesToVisit;

    if (!source) {
      return [];
    }

    const sourceUris = importModuleGraph.get(source.uri) ?? [];
    const importingModules = sourceUris.filter(
      (importingModule, index, importers) =>
        !visited.has(importingModule.uri) &&
        importers.findIndex((module) => module.uri === importingModule.uri) ===
          index,
    );
    const nextVisited = new Set([
      ...visited,
      ...importingModules.map((module) => module.uri),
    ]);

    return [
      ...importingModules,
      ...findTransitiveImporters(
        [...remainingModules, ...importingModules],
        nextVisited,
      ),
    ];
  };

  return findTransitiveImporters([sourceFile], new Set([sourceFile.uri]));
}
```

### test

`elm-language-server/test/typeCheckerImportGraph.performance.test.ts`

```ts
import { performance } from "perf_hooks";
import { getTransitiveImportingModules } from "../src/compiler/typeChecker";

interface IStressModule {
  uri: string;
}

type ImportModuleGraph = ReadonlyMap<string, readonly IStressModule[]>;

const NUMBER_OF_LAYERS = 24;
const MODULES_PER_LAYER = 24;
const WARMUP_ITERATIONS = 5;
const BENCHMARK_ITERATIONS = 20;
const EXPECTED_MODULES = NUMBER_OF_LAYERS * MODULES_PER_LAYER;

function getTransitiveImportingModulesMutable(
  importModuleGraph: ImportModuleGraph,
  sourceFile: IStressModule,
): IStressModule[] {
  const importingModules: IStressModule[] = [];
  const visited = new Set<string>([sourceFile.uri]);
  const modulesToVisit = [sourceFile];

  for (let index = 0; index < modulesToVisit.length; index++) {
    const source = modulesToVisit[index];

    (importModuleGraph.get(source.uri) ?? []).forEach((importingModule) => {
      if (!visited.has(importingModule.uri)) {
        visited.add(importingModule.uri);
        importingModules.push(importingModule);
        modulesToVisit.push(importingModule);
      }
    });
  }

  return importingModules;
}

function createStressGraph(): {
  importModuleGraph: Map<string, IStressModule[]>;
  sourceFile: IStressModule;
} {
  const sourceFile = { uri: "Root" };
  const layers = Array.from({ length: NUMBER_OF_LAYERS }, (_, layer) =>
    Array.from({ length: MODULES_PER_LAYER }, (_, module) => ({
      uri: `Layer${layer}.Module${module}`,
    })),
  );
  const importModuleGraph = new Map<string, IStressModule[]>();

  importModuleGraph.set(sourceFile.uri, layers[0]);

  layers.forEach((layer, layerIndex) => {
    const nextLayer = layers[layerIndex + 1];

    layer.forEach((module, moduleIndex) => {
      if (!nextLayer) {
        // Exercise cycle handling and duplicate edges at the end of the graph.
        importModuleGraph.set(module.uri, [sourceFile, sourceFile]);
        return;
      }

      const adjacentModule = nextLayer[(moduleIndex + 1) % MODULES_PER_LAYER];
      const distantModule =
        nextLayer[(moduleIndex + MODULES_PER_LAYER / 2) % MODULES_PER_LAYER];
      const sameModule = nextLayer[moduleIndex];

      importModuleGraph.set(module.uri, [
        sameModule,
        adjacentModule,
        distantModule,
        sameModule,
      ]);
    });
  });

  return { importModuleGraph, sourceFile };
}

function measure(
  operation: () => IStressModule[],
  iterations: number,
): { elapsed: number; visitedModules: number } {
  const start = performance.now();
  let visitedModules = 0;

  for (let iteration = 0; iteration < iterations; iteration++) {
    visitedModules += operation().length;
  }

  return { elapsed: performance.now() - start, visitedModules };
}

describe("transitive importing module traversal performance", () => {
  it("stress tests the immutable traversal against the previous mutable loop", () => {
    const { importModuleGraph, sourceFile } = createStressGraph();
    const mutableTraversal = (): IStressModule[] =>
      getTransitiveImportingModulesMutable(importModuleGraph, sourceFile);
    const immutableTraversal = (): IStressModule[] =>
      getTransitiveImportingModules(importModuleGraph, sourceFile);

    const expected = mutableTraversal();
    expect(expected).toHaveLength(EXPECTED_MODULES);
    expect(immutableTraversal()).toEqual(expected);

    measure(mutableTraversal, WARMUP_ITERATIONS);
    measure(immutableTraversal, WARMUP_ITERATIONS);

    const mutable = measure(mutableTraversal, BENCHMARK_ITERATIONS);
    const immutable = measure(immutableTraversal, BENCHMARK_ITERATIONS);

    expect(mutable.visitedModules).toBe(
      EXPECTED_MODULES * BENCHMARK_ITERATIONS,
    );
    expect(immutable.visitedModules).toBe(
      EXPECTED_MODULES * BENCHMARK_ITERATIONS,
    );

    console.info(
      `Mutable: ${mutable.elapsed.toFixed(2)}ms; immutable: ${immutable.elapsed.toFixed(2)}ms`,
    );
    expect(mutable.elapsed).toBeLessThan(immutable.elapsed);
  });
});
```

jasonatwell: I asked the LLM to speed up the functional implementation and this is what it came up with, averages under 7ms on that same test. While about 10x slower than the mutable version, sub 7ms seems reasonable. I leave it up to maintainers to decide what to do. I'll test out this functional implementation and if I run into any problems I'll update here

```ts
export function getTransitiveImportingModules<T extends { uri: string }>(
  importModuleGraph: ReadonlyMap<string, readonly T[]>,
  sourceFile: T,
): T[] {
  const findTransitiveImporters = (
    modulesToVisit: readonly T[],
    visited: ReadonlySet<string>,
  ): T[] => {
    const candidateModules = modulesToVisit.flatMap(
      (source) => importModuleGraph.get(source.uri) ?? [],
    );
    const importingModules = Array.from(
      new Set(candidateModules.filter((module) => !visited.has(module.uri))),
    );

    if (importingModules.length === 0) {
      return [];
    }

    const nextVisited = new Set([
      ...visited,
      ...importingModules.map((module) => module.uri),
    ]);

    return [
      ...importingModules,
      ...findTransitiveImporters(importingModules, nextVisited),
    ];
  };

  return findTransitiveImporters([sourceFile], new Set([sourceFile.uri]));
}
```

test/typeCheckerImportGraph.performance.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+interface IStressModule {
+  uri: string;
+}
+
+const NUMBER_OF_LAYERS = 24;
+const MODULES_PER_LAYER = 24;
+const WARMUP_ITERATIONS = 5;
+const BENCHMARK_ITERATIONS = 20;
+const FUNCTIONAL_TARGET_MS = 10;
```
jasonatwell: The first mutable version was able to consistently pass this test in under 1ms, so I felt that a functional version that's only 10x slower would be sufficient, which is how I can to 10ms as a target. I tested this on an alpine image with 2 cpus and 512M memory and was able to consistently pass these tests. With a single cpu the test fails, taking about 25ms

```
container run --rm \
  --cpus 2 \
  --memory 512M \
  --mount type=bind,source="$PWD",target=/src,readonly \
  --entrypoint sh \
  node:22-alpine -lc '
    mkdir -p /work
    tar --exclude=node_modules --exclude=.git -C /src -cf - . | tar -C /work -xf -
    cd /work
    apk add --no-cache python3 make g++
    npm ci

    for run in $(seq 1 10); do
      npx jest --runInBand --coverage=false \
        test/typeCheckerImportGraph.performance.test.ts
    done
  '
```

jasonatwell:
Tests failed because they didn't meet the 10ms limit. [windows latest 22](https://github.com/elm-tooling/elm-language-server/actions/runs/31845938719/job/94994630553) was the slowest at 20.47ms.

If we want to enforce a fast solution we should bump this up to 30ms to account for variance, but if we just want to ensure this works in a reasonable amount of time, the threshold should probably be closer to 100ms so this test never fails unless there's a significant performance degradation. Personally I'd prefer to keep it fast, 30-50ms, but since I don't have to deal with the annoyance of randomly failing tests I think my vote counts 0% 😆

@razzeee what do you think?

razzeee:
would the non functional way fix windows?

jasonatwell:
> would the non functional way fix windows?

I assume so, since the non-functional implementation takes less than 1ms on my machine, but I don't have a good way of testing on windows locally

razzeee:
might be worth it then, can you revert it?

jasonatwell:
With the mutable code, [windows 22, latest](https://github.com/elm-tooling/elm-language-server/actions/runs/32079939968/job/95608973063#step:4:447) still failed but it took 15ms instead of 20.47. I think we should bump up the target to 30ms and keep the mutable code.

razzeee:
Thanks for your patience

